PR #2, opened 2023-01-24
Helper function to read h5ad file from google bucket

ordabayevy:


scvid/data/read.py
```diff
@@ -0,0 +1,43 @@
...
+
+    Args:
+        filename (str): Path to the data file in Cloud Storage.
+    """
+    assert filename.startswith("gs:")
+    # parse bucket and blob names from the filename
+    filename = re.sub(r"^gs://?", "", filename)
+    bucket_name, blob_name = filename.split("/", 1)
+
+    storage_client = storage.Client()
```
mbabadi, line 23: Consider the following change:

```
from google.cloud.storage import Client

def read_h5ad_gcs(filename: str, storage_client: Optional[Client]):
    ...
    if storage_clientis None:
        storage_client= Client()
    ... 
```

We will soon find use-cases where the GCS storage client needs to be authenticated with privileged API tokens.

scvid/data/read.py
```diff
@@ -0,0 +1,43 @@
...
+    bucket = storage_client.bucket(bucket_name)
+    blob = bucket.blob(blob_name)
+
+    with blob.open("rb") as f:
+        return read_h5ad(f)
+
+
+def read_h5ad_filename(filename: str) -> AnnData:
```
mbabadi, line 31: Consider further refactoring to `read_h5ad_local` and `read_h5ad` which handles mixed input.

mbabadi (changes requested):

ordabayevy:
@mbabadi does it look good now?

mbabadi:
@ordabayevy Almost :) The reason I suggested declaring the kwarg `storage_client` as `Optional[Client] = None`  was due to python's eagerness to evaluate all expressions, and that includes kwarg default values. Making a habit of declaring _expensive_ kwarg default values (such as `Client()`) is one reason some python package are slow to import. I personally avoid this pattern, but that's entirely up to you.

ordabayevy:
@mbabadi interesting, good to know! Actually glad I wrote it that way first so now I know more about Python :)

mbabadi (approved):

PR #3, opened 2023-01-25
`AnnDataSchema` for anndata validation

ordabayevy:


ordabayevy:
@mbabadi can you have a look if this validation method looks right?

mbabadi:
@ordabayevy looking at it right now.

scvid/data/schema.py
```diff
@@ -0,0 +1,70 @@
...
+    Args:
+        adata (AnnData): Reference AnnData object.
+    """
+
+    attrs = ["obs", "obsm", "var", "varm", "var_names", "layers"]
+
+    def __init__(self, adata: AnnData) -> None:
+        self.attr_values = {}
+        for attr in self.attrs:
+            self.attr_values[attr] = getattr(adata, attr)
```
mbabadi, line 31: Can you check that `del adata` does not invalidate references to `self.attr_values`?

scvid/data/schema.py
```diff
@@ -0,0 +1,70 @@
...
+        >>> adata.obs["batch"] = [2, 3, 4, 5]
+        >>> adata.var["mu"] = ["a", "b", "c"]
+        >>> schema = AnnDataSchema(ref_adata)
+        >>> schema.validate_adata(adata)
+
+    Args:
+        adata (AnnData): Reference AnnData object.
+    """
+
+    attrs = ["obs", "obsm", "var", "varm", "var_names", "layers"]
```
mbabadi, line 26: Please add `varp` attr too. It is a dictionary of square `n_var x n_var` matrices. It should identically match.

test/test_data.py
```diff
@@ -0,0 +1,57 @@
...
+        dtype=X.dtype,
+        obs=obs,
+        var=var,
+        layers={"L": L},
+        obsm={"M": M},
+        varm={"V": V},
+    )
+
+
+def test_validate_adata(ref_adata, adata):
```
mbabadi, line 55: What do you think about adding a couple of failing fixtures and test them for failure, for example:
- `obs` columns permuted
- `obs` column names are different
- `obs` columns names match but dtype of a column is different
- `obs` columns names match but the categories of a `pandas.Categorical` column are different
- `var` column names are in different order

Testing for failure may seem a bit redundant if the `equals` calls behave as expected, but we don't know that. WDYT?

mbabadi (changes requested):

scvid/data/schema.py
ordabayevy: Done.

ordabayevy (commented):
@mbabadi Added additional tests that you recommended.

test/test_data.py
ordabayevy: I agree, it makes sense to add tests for failure.

mbabadi (approved):

ordabayevy:
Thanks for reviewing @mbabadi !

PR #6, opened 2023-02-07
Probabilistic PCA

ordabayevy:
Probabilistic PCA implemented in two versions:
- with z marginalized out (`marginalize_z=True` default flag) - seems to work better
- with z not marginalized out (`marginalize_z=False` flag)

Added a test that checks explained variance and that inferred `z` has std close to 1 for a test data.

### DIFF
--- a/scvid/module/_probabilisticpca.py
+++ b/scvid/module/_probabilisticpca.py
@@ -17,7 +17,7 @@ class ProbabilisticPCAPyroModel(PyroModule):
         n_vars: Number of input features.
         n_components: Number of components to model.
         mean: Mean of the input data.
-        marginalize_z: Marginalize latent variable z.
+        marginalize_z: Marginalize out latent variable z.
     """
 
     def __init__(
@@ -79,7 +79,7 @@ class ProbabilisticPCAPyroGuide(PyroModule):
         n_vars: Number of input features.
         n_components: Number of components to model.
         mean: Mean of the input data.
-        marginalize_z: Marginalize latent variable z.
+        marginalize_z: Marginalize out latent variable z.
     """
 
     def __init__(
@@ -122,7 +122,7 @@ class ProbabilisticPCAPyroModule(PyroBaseModuleClass):
         n_vars: Number of input features.
         n_components: Number of components to model.
         mean: Mean of the input data.
-        marginalize_z: Marginalize latent variable z.
+        marginalize_z: Marginalize out latent variable z.
     """
 
     def __init__(
--- a/scvid/module/_probabilisticpca.py
+++ b/scvid/module/_probabilisticpca.py
@@ -164,6 +164,12 @@ class ProbabilisticPCAPyroModule(PyroBaseModuleClass):
         """
         Return the latent representation for each cell.
         """
-        L = self._guide.L
+        if self.marginalize_z:
+            W = self._model.W
+            sigma = self._model.sigma
+            M = W @ W.T + sigma**2 * torch.eye(len(W))
+            L = W.T @ torch.linalg.inv(M)
+        else:
+            L = self._guide.L
         z_loc = (x - self.mean) @ L
         return z_loc
--- a/scvid/module/_probabilisticpca.py
+++ b/scvid/module/_probabilisticpca.py
@@ -1,0 +1,169 @@
+import pyro
+import pyro.distributions as dist
+import torch
+from pyro.nn import PyroModule, PyroParam
+from scvi.module.base import PyroBaseModuleClass
+from torch.distributions import constraints
+
+_PROBABILISTIC_PCA_PYRO_MODULE_NAME = "probabilistic_pca"
+
+
+class ProbabilisticPCAPyroModel(PyroModule):
+    """
+    A PyroModule that serves as the model for the ProbabilisticPCAPyroModule class.
+
+    Args:
+        n_obs: Number of observations.
+        n_vars: Number of input features.
+        n_components: Number of components to model.
+        mean: Mean of the input data.
+        marginalize_z: Marginalize latent variable z.
+    """
+
+    def __init__(
+        self,
+        n_obs: int,
+        n_vars: int,
+        n_components: int,
+        mean: torch.Tensor,
+        marginalize_z: bool,
+    ):
+        super().__init__(_PROBABILISTIC_PCA_PYRO_MODULE_NAME)
+
+        self.n_obs = n_obs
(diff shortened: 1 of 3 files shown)


scvid/module/_probabilisticpca.py
```diff
@@ -0,0 +1,175 @@
...
+        super().__init__(_PROBABILISTIC_PCA_PYRO_MODULE_NAME)
+
+        self.n_obs = n_obs
+        self.n_vars = n_vars
+        self.n_components = n_components
+        self.mean = mean
+        self.marginalize_z = marginalize_z
+
+        # model parameters
+        self.W = PyroParam(lambda: torch.randn((n_components, n_vars)))
```
mbabadi, line 40: Please let the scale of the random initialization of `W` (and every other model parameter) to be user-specified. There are often good heuristics to guess the _magnitude_ of various model params, and so you'd want to expose these to the user. For example, let's say we know the empirical variance of the dataset, and let that be `s^2`. A PPCA model with Gaussian initialization of `W` witth scale `w` produces the following variance:
```
E[tr(W^T W) + F \sigma^2] = F D w^2 + F \sigma^2 ~ s^2
```
where `F` is the ambient dimension = number of features, and we have `D` PCs. Now, let's say we expect to be able to explain `90%` of the variance with the PCs and leave `10%` for the isotropic noise. The we have:
```
\sigma^2 ~ 0.1 s^2 / F
w^2 ~ 0.9 s^2 / (F x D)
```

scvid/module/_probabilisticpca.py
```diff
@@ -0,0 +1,175 @@
...
+        self.sigma = PyroParam(
+            lambda: torch.tensor(1.0), constraint=constraints.positive
```
mbabadi, line 42: (expose the default value to the initializer -- see above)

scvid/module/_probabilisticpca.py
```diff
@@ -0,0 +1,175 @@
...
+        self.n_obs = n_obs
+        self.n_vars = n_vars
+        self.n_components = n_components
+        self.mean = mean
+        self.marginalize_z = marginalize_z
+
+        # guide parameters
+        if not self.marginalize_z:
+            self.L = PyroParam(lambda: torch.randn((n_vars, n_components)))
+            self.z_scale = PyroParam(
```
mbabadi, line 104: Just FYI, this guide is not powerful enough to represent the exact solution, see Eq. (6) here:
http://www.cs.columbia.edu/~blei/seminar/2020-representation/readings/TippingBishop1999.pdf

The exact posterior implies a multivariate normal for `z | x` with the data-independent covariance matrix. Perhaps implement that as well and let the user select the guide type (e.g. `diagonal_normal` and `multivariate_normal`)? Since we expect to work with < 200 PCs in practice, a MVN guide for `z`.

Out of curiosity, did you implement the exact guide from Bishop's paper at some point? Was it too slow?

test/test_ppca.py
```diff
@@ -0,0 +1,59 @@
...
+@pytest.mark.parametrize("minibatch", [False, True], ids=["fullbatch", "minibatch"])
+def test_probabilistic_pca(x_np, minibatch, marginalize):
+    x_mean = x_np.mean(axis=0)
+
+    pyro.clear_param_store()
+    ppca = ProbabilisticPCAPyroModule(
+        n_obs=n, n_vars=p, n_components=k, mean=x_mean, marginalize_z=marginalize
+    )
+    elbo = infer.Trace_ELBO()
+    adam = optim.Adam({"lr": 1e-2})
```
mbabadi, line 35: Good test!

(unrelated to this PR -- but I have a feeling that scheduling the LR e.g. w/ the usual cosine annealing might do wonders with PPCA. why? because it's a simple model w/ a convex loss and after the initial convergence to a rough solution, annealing can let the "ball" roll down to the global minimum. It'd be great to try this out in your experiments!).

scvid/module/_probabilisticpca.py
ordabayevy: Yeah, the exact guide seems to be much slower.

ordabayevy:
@mbabadi thank for the great feedback! I addressed the comments. I left out the exact solution for the guide as it seems to be slow even on the toy data. But I haven't done comprehensive benchmarking, maybe can be done in the future.
Also I don't understand why do we need static RNG with a specified seed. Can you explain that more? If I add it does it have to be provided by the user (with a default value)?

mbabadi (changes requested):
Thanks for incorporating my suggestions! :-)

As for using static RNG (by which I mean doing `rng = torch.Generator(device=...); rng.manual_seed(<seed>); rng.randn(...)` rather than messing around with the PyTorch's global RNG state, i.e. `torch.manual_seed(<seed>);  torch.randn(...)`): it is to eliminate side effects, that's all. In general, adherence to functional programming and writing "pure functions" and "pure classes" is a good strategy to guard against future headaches.

scvid/module/_probabilisticpca.py
```diff
@@ -0,0 +1,150 @@
...
+            assert mean_g.shape == (
+                g_genes,
+            ), "Expected meang_g to have a shape ({g_genes},) but found {mean_g.shape}."
+        if mean_g is None:
+            # make mean_g a learnable parameter
+            self.mean_g = PyroParam(lambda: torch.zeros(g_genes))
+        else:
+            self.mean_g = mean_g
+
+        rng = torch.Generator()
```
ordabayevy, line 55: Not sure how to handle the `device` at the moment since I don't understand well how it is managed by the PyTorch Lightning. If any issues will come up with the device maybe we can address it then?

mbabadi: Sounds good! The [LightningModule](https://pytorch-lightning.readthedocs.io/en/stable/api/pytorch_lightning.core.LightningModule.html#pytorch_lightning.core.LightningModule) knows what device it is on. You can access the reference via `self.device`. A vanilla `nn.Module`, however, does not have `self.device` ...

ordabayevy: I see. Can you approve this PR if everything looks ok to you?

mbabadi (approved):

scvid/module/_probabilisticpca.py
mbabadi: Done! ✅

ordabayevy: Yay!

PR #13, opened 2023-02-14
Distributed Training POC

kcibul:
Adds a DistributedSampler that is aware of the chunk structure to enable efficient multi-process/gpu training.

Example code for running distributed training using this approach, as well as one using distributed sampling built into a native dataset for AnnData is included


ordabayevy:
Closed in favor of #21.

Issue #11, opened 2023-02-10
scRNA-seq data normalization for PCA


mbabadi:
PCA expects Gaussian distributed data, whereas raw scRNA-seq data is integral. The heuristic in the field is to apply the following transformations:

$\ell_n = \sum_g X_{ng}$
$y_{ng} = \log(1 + C \times X_{ng} / \ell_n)$
$z_{ng} = (y_{ng} - \mu_g) / \sigma_g$

where $C = 10,000$ is a "pseudocount", and:

$\mu_g = \frac{1}{N}\sum_n y_{ng}$
$\sigma^2_g = \frac{1}{N}\sum_n (y_{ng} - \mu_g)^2$

Feature whitening, i.e. dividing by $\sigma_g$, is considered "optional" and is sometimes skipped (reasoning: a more variable gene is likely to be a more biologically informative gene, and we should not throw out that information).

This normalization recipe introduces two problems:
- The recipe is an integral part of the PCA transformation of the data, and applying a learned PCA transformation to user's data requires the knowledge of the parameters of this normalization recipe: $C$, $\mu_g$, and $\sigma_g$.
- $\mu_g$ and $\sigma_g$ must be calculated from full batch data

The first problem is easy to solve:  we create a `LogNormalizationTransform` module which is passed the PCA module. The parameters of the transformation naturally become part of the state of the PCA module.

The second problem, i.e estimation of full batch $\mu_g$ and $\sigma_g$, can be done in 3 different ways:
- A single-pass estimation using BigQuery; $\mu_g$ and $\sigma_g$ would then be written to the `.var` metadata, and `C` to the `.uns` dictionary.
- A single-pass estimation from AnnData chunks; this can be parallelized, but will be inevitably slower than the previous approach and add one extra epoch of data loading.
- Online estimation: we can use ideas from `BatchNorm` (and `SyncBatchNorm` for parallelization) to estimate $\mu_g$ and $\sigma_g$ from minibatches; we can also increase the exponential moving average (EMA) window as training proceeds to stabilize the estimates.

ordabayevy:
I added the `LogNormalize` transform in #15. @mbabadi isn't it a bit redundant that we have the mean parameter both in `LogNormalize` and `ProbabilisticPCA`?

PR #15, opened 2023-02-15
`LogNormalize` transform

ordabayevy:
This PR add `LogNormalize` transform described in #11

### DIFF
--- a/scvid/transforms/transforms.py
+++ b/scvid/transforms/transforms.py
@@ -1,0 +1,28 @@
+import torch
+from torch import nn
+
+
+class LogNormalize(nn.Module):
+    """
+    LogNormalize gene counts with mean, standard deviation, and pseudo-count.
+
+    Args:
+        mean_g: Means for each gene.
+        std_g: Standard deviations for each gene.
+        C: Gene epxression pseudo-count.
+    """
+
+    def __init__(self, mean_g: torch.Tensor, std_g: torch.Tensor, C: int = 10_000):
+        super().__init__()
+        self.mean_g = mean_g
+        self.std_g = std_g
+        self.C = C
+
+    def forward(self, x_ng: torch.Tensor) -> torch.Tensor:
+        l_n1 = x_ng.sum(axis=-1, keepdim=True)
+        y_ng = torch.log1p(self.C * x_ng / l_n1)
+        z_ng = (y_ng - self.mean_g) / self.std_g
+        return z_ng
+
+    def __repr__(self) -> str:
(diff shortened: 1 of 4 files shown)


scvid/transforms/transforms.py
```diff
@@ -0,0 +1,28 @@
+import torch
+from torch import nn
+
+
+class LogNormalize(nn.Module):
```
mbabadi, line 5: Let's rename this to something more appropriate, like `ZScoreLogNormalize`.

scvid/transforms/transforms.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    LogNormalize gene counts with mean, standard deviation, and pseudo-count.
```
mbabadi, line 7: I belive I was sloppy with me description.  In the following expression:
$log(1 + C \times x_{ng} / \sum_g x_{ng})$

The pseudo-count is 1, and C, is fixed target total count (not a pseudo-count). Let's do the following:

- Rename `C` to `target_count`
- Make scaling by `std_g` optional via a `scale` boolean flag.
- If applicable, add `eps: float = 1e-6` (a new __init__ arg) to `std_g` in the denominator to obviate division by zero.
-

scvid/transforms/transforms.py
```diff
@@ -0,0 +1,28 @@
...
+
+    Args:
+        mean_g: Means for each gene.
+        std_g: Standard deviations for each gene.
+        C: Gene epxression pseudo-count.
+    """
+
+    def __init__(self, mean_g: torch.Tensor, std_g: torch.Tensor, C: int = 10_000):
```
mbabadi, line 15: (see above)

mbabadi (changes requested):

scvid/transforms/transforms.py
mbabadi: or better yet,`ZScoreLog1pNormalize`.

scvid/transforms/transforms.py
ordabayevy: Or maybe have `std_g` as `Optional[torch.Tensor] = None` and not apply it if it is None?

ordabayevy:
@mbabadi ready for a review!

mbabadi (approved):

mbabadi:
@ordabayevy oh wait

mbabadi:
@ordabayevy now that you're at it, how about also safeguarding dividing by library size (+ eps)? we don't want a NaN if a user gives us an empty cell :D also, I am leading toward adding `perform_scaling` bool.  It will be a stronger commitment from the user. Always a good thing!

mbabadi:
Great! Feel free to merge anytime! :)

PR #19, opened 2023-02-28
`IterableDistributedAnnDataCollectionDataset`

ordabayevy:


ordabayevy:
@mbabadi can you review this? I came up with a cool trick to get cache and worker info for in unit tests :)

scvid/data/dadc_dataset.py
```diff
@@ -14,11 +16,161 @@ def __init__(self, dadc: DistributedAnnDataCollection) -> None:
     def __len__(self) -> int:
         return len(self.dadc)
 
-    def __getitem__(self, idx: int) -> Dict[str, Tensor]:
+    def __getitem__(self, idx: int) -> Dict[str, np.ndarray]:
         """Return gene counts for a cell at idx."""
```
mbabadi, line 20: `gene counts` -> `feature counts`. Also, mention that if the data is in sparse repr, it will be made dense.

scvid/data/dadc_dataset.py
```diff
@@ -14,11 +16,161 @@ def __init__(self, dadc: DistributedAnnDataCollection) -> None:
...
+        # for testing purposes
+        if self.test_mode:
+            worker_info = get_worker_info()
+            if worker_info is not None:
+                data["worker_id"] = np.array([worker_info.id])
+            data["miss_count"] = np.array([self.dadc.cache.miss_count])
+
+        return data
+
+    def __iter__(self):
```
mbabadi, line 83: Please extend the Docstring: that we use `torch.utils.data.get_worker_info()` to determine the invocation context. If single worker, then we iterate over the entire datasets. Otherwise, we iterate over a subset of cells in a manner that minimizes the overlap between the data chunks loaded by each worker.

mbabadi (changes requested):

scvid/data/dadc_dataset.py
mbabadi: @ordabayevy solid work!

What's the plan for multi-GPU? One approach is to create separate DADCs for each worker based on their rank. Another approach (perhaps cleaner) would be to get the world size and node rank from `torch.distributed` and treat `work_size x num_workers` as `total_workers`. The same logic that you've implemented should work simply by replacing `num_workers` with `total_workers`.

I would also suggest factoring out the code that generate `indices` out of `__iter__`. Then you can make an issue to make a smarter assignment strategy where you balance the number of cells given to each worker using a greedy algorithm (this is low priority for our current use case).

ordabayevy: Thanks for reviewing @mbabadi, addressed your comments. I have an implemented version for the multi-GPU case here https://github.com/cellarium-ai/scvi-distributed/pull/21/files.
Working on tests at the moment :) We can discuss it tomorrow at our meeting.

mbabadi (approved):

ordabayevy:
Thanks @mbabadi !

PR #28, opened 2023-03-08
Fix anndata and scvi-tools dependencies

### LINKED ISSUE #26: Switch dependency on `anndata` and `scvi-tools` from our forks to original latest versions.
We actually don't need any changes in `anndata` and `scvi-tools` libraries for `scvi-distributed` so we can switch dependency to their latest versions from our forks.

ordabayevy:
Closes #26.

### DIFF
--- a/scvid/data/distributed_anndata.py
+++ b/scvid/data/distributed_anndata.py
@@ -1,7 +1,6 @@
 from contextlib import contextmanager
 from typing import List, Optional, Sequence, Tuple, Union
 
-import numpy as np
 import pandas as pd
 from anndata import AnnData
 from anndata._core.index import Index, _normalize_indices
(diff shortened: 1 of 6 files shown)


scvid/data/distributed_anndata.py
```diff
@@ -307,12 +305,3 @@ def __repr__(self) -> str:
                 if len(keys) > 0:
                     descr += f"\n    {attr}: {str(list(keys))[1:-1]}"
         return descr
-
-
-@get_anndata_attribute.register
```
mbabadi, line 312: For my education -- what happened here? why removed it? :)

mbabadi (approved):

scvid/data/distributed_anndata.py
ordabayevy: This is the function that is used by `AnnTorchDataset` in `scvi-tools`. In our fork I made it single-dispatch and registered its implementation for `AnnCollection`. Since we don't use `scvi-tools` training machinery we don't need this.

PR #31, opened 2023-03-09
Garbage collect anndata

### LINKED ISSUE #23: Fix AnnData memory leak issue
AnnData is not properly garbage collected which leads to CPU memory overflows when training models on large datasets. See https://github.com/scverse/anndata/issues/360

ordabayevy:
Fixes #23.

### DIFF
--- a/scvid/data/distributed_anndata.py
+++ b/scvid/data/distributed_anndata.py
@@ -1,6 +1,7 @@
 # Copyright Contributors to the Cellarium project.
 # SPDX-License-Identifier: BSD-3-Clause
 
+import gc
 from contextlib import contextmanager
 from typing import List, Optional, Sequence, Tuple, Union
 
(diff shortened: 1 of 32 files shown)


ordabayevy:
@mbabadi can you review this? It would be awesome if you had tests ideas for anndata garbage collection.

scvid/data/distributed_anndata.py
```diff
@@ -301,7 +302,14 @@ def adata(self) -> AnnData:
             )
             self.schema.validate_anndata(adata)
             # cache anndata
-            self.cache[self.filename] = adata
+            if len(self.cache) < self.cache.max_size:
+                self.cache[self.filename] = adata
+            else:
+                self.cache[self.filename] = adata
```
mbabadi, line 308: This looks alright to me, but I'd suggest having an eye on memory leak in real usage (training on lots of AnnData) in case the `gc.collect()` trick is not sufficient. There were other tricks, too (e.g. replacing `.X`. with an empty sparse matrix _before_ calling collect() -- sounds magical to me but the guy in the scverse issue claimed it did the trick).

mbabadi (approved):

scvid/data/distributed_anndata.py
ordabayevy: I'll keep an eye on it. If there is a problem it should quickly become apparent once we start analyzing much larger datasets. I just don't want to introduce unnecessary fixes before an actual issue arises.

PR #42, opened 2023-03-21
Add public anndata test files

### LINKED ISSUE #22: Create public test anndata
Create public anndata files in GCS that can be used for testing.

ordabayevy:
Closes #22.

### DIFF
--- a/test/test_examples.py
+++ b/test/test_examples.py
@@ -14,20 +14,18 @@ EXAMPLES = [
     (
         "probabilistic_pca.py "
         "--filenames https://storage.googleapis.com/dsp-cellarium-cas-public/test-data/benchmark_v1.{000..001}.h5ad "
-        "--batch_size 5000 --accelerator cpu --max_steps 4 --strategy ddp"
+        "--batch_size 5000 --accelerator cpu --max_steps 4 --devices 1"
     ),
     (
         "onepass_mean_var_std.py "
         "--filenames https://storage.googleapis.com/dsp-cellarium-cas-public/test-data/benchmark_v1.{000..001}.h5ad "
-        " --batch_size 5000 --accelerator cpu --strategy ddp"
+        " --batch_size 5000 --accelerator cpu --devices 1"
     ),
 ]
 
 
 @pytest.mark.parametrize("example", EXAMPLES)
-def test_cpu_multi_device(example):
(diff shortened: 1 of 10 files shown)


test/test_examples.py
```diff
@@ -11,16 +11,21 @@
...
+    ),
 ]
 
 
-@pytest.mark.skipif("CI" in os.environ, reason="GCS keys are not available")
 @pytest.mark.parametrize("example", EXAMPLES)
-def test_cpu_multi_device(example):
-    devices = os.environ.get("TEST_DEVICES", "1")
-    example += f" --devices {devices}"
+def test_cpu(example):
```
ordabayevy, line 28: Removed `_multi_device` because of the error below (with `TEST_DEVICES=2`) which I could not get fixed

```
RuntimeError: The server socket has failed to listen on any local network address. The server socket has failed to bind to [::]:42063 (errno: 98 - Address already in use). The server socket has failed to bind to 0.0.0.0:42063 (errno: 98 - Address already in use).
```
Strangely this only happens when running all multi device tests (`pytest -k multi_device`) but doesn't happen when I run only `test_examples` (`pytest test/test_examaples.py`).

test/test_onepass_mean_var_std.py
```diff
@@ -117,7 +117,6 @@ def test_onepass_mean_var_std_iterable_dataset_multi_device(
         accelerator="cpu",
         devices=devices,
         max_epochs=1,  # one pass
-        strategy="ddp",
```
ordabayevy, line 120: `ddp` is automatically selected by PyTorch Lightning 2.0

test/test_examples.py
mbabadi: I have a hunch that `ddp_spawn` will iron out these issues (though it requires some modifications because the model on the main process will not be updated, as we discussed).  Perhaps give it a try, but we can totally live with just testing `TEST_DEVICES=2` for the time being (so totally feel free to ignore).

mbabadi (approved):

test/test_examples.py
ordabayevy: Just tried it with `ddp_spawn` but still get the same error message. I wonder if it has to do with using `check_call([sys.executable, filename] + args)`.

ordabayevy:
I created a new issue #44 and merge this PR for now.

PR #43, opened 2023-03-23
Use analytic solutions for non-marginalized PPCA guides

ordabayevy:


scvid/module/_probabilisticpca.py
```diff
@@ -66,28 +78,19 @@ def __init__(
...
         # guide parameters
-        if ppca_flavor == "marginalized":
-            pass
-        elif ppca_flavor == "diagonal_normal":
-            self.L_gk = PyroParam(
-                lambda: torch.randn((g_genes, k_components), generator=rng)
-            )
+        if ppca_flavor == "diagonal_normal":
+            M_inv_kk = torch.linalg.inv(self.M_kk)
+            z_scale_k_init = torch.sqrt(torch.diag(self.sigma**2 * M_inv_kk)).detach()
```
mbabadi, line 83: I don't think taking the diagonal entries of a potentially correlated covariance matrix is necessarily a good initialization, but I don't have a better suggestion either :)  Here's the thing though:  if `W_kg` is initialized randomly and `G >> K` , then `M_kk` is diagonal to a decent approximation.

mbabadi: So we're probably OK here.

scvid/module/_probabilisticpca.py
```diff
@@ -123,22 +126,30 @@ def model(self, x_ng: torch.Tensor):
...
+            z_loc_nk = torch.linalg.solve(self.M_kk, WX_kn).T
 
             if self.ppca_flavor == "diagonal_normal":
                 pyro.sample("z", dist.Normal(z_loc_nk, self.z_scale_k).to_event(1))
             elif self.ppca_flavor == "multivariate_normal":
+                M_inv_kk = torch.linalg.inv(self.M_kk)
+                # This function skips the (slow) error checking and error message construction
+                # of torch.linalg.cholesky(), instead directly returning the LAPACK error codes
+                # as part of a named tuple (L, info).
+                z_scale_tril_kk, _ = torch.linalg.cholesky_ex(
```
mbabadi, line 147: Nice, I wasn't aware of this! :)

scvid/module/_probabilisticpca.py
```diff
@@ -123,22 +126,30 @@ def model(self, x_ng: torch.Tensor):
...
+                    self.sigma**2 * M_inv_kk
+                )
```
mbabadi, line 149: Consider trying out the following stabilizing tricks in case you still run into NaNs:

(1) Try `self.sigma**2 * M_inv_kk + self.jitter * torch.eye(self.k_components, device=self.sigma.device)` as the Cholesky argument;

(2) Try `z_scale_tril_kk.view(-1)[::(self.k_components + 1)] += self.jitter` post Cholesky.

(3) Add the jitter directly to `W_kg @ W_kg.T` in `self.M_kk`.

Set `self.jitter` to 1e-6 initially, and potentially increase to 1e-3.

scvid/module/_probabilisticpca.py
```diff
@@ -153,16 +164,12 @@ def get_latent_representation(
...
-        z_loc_nk = (x_ng - self.mean_g) @ L_gk
+        WX_kn = self.W_kg @ (x_ng - self.mean_g).T
+        z_loc_nk = torch.linalg.solve(self.M_kk, WX_kn).T
         return z_loc_nk
+
+    @torch.inference_mode()
+    def log(self, plan: pl.LightningModule) -> None:
+        """Logging to TensorBoard by default"""
+        var_explained = torch.sum(torch.diag(self.W_kg.T @ self.W_kg) + self.sigma**2)
+        plan.log("var_explained", var_explained)
```
mbabadi, line 175: This is good, but also consider calculating the total variance (from one-pass-mean-std result) as `self.total_variance` as a cached property and log the ratios (see below).  If you want to make your code agonistic to the presence of one-pass-mean-std results, this class can take a `total_variance` optional callable. If it is not None, then log the ratios as well as the unnormalized var_explained.

```
var_explained_W = torch.trace(self.W_kg.T @ self.W_kg)
plan.log("var_explained", self.sigma**2 + var_explained_W )
plan.log("var_explained_W",  var_explained_W)
plan.log("var_explained_sigma", self.sigma**2)
if self.total_variance is not None:
    plan.log("var_explained_ratio", (var_explained_W + self.sigma**2)/ self.total_variance)
    plan.log("var_explained_ratio_W",  var_explained_W / self.total_variance)
    plan.log("var_explained_ratio_sigma", self.sigma**2 / self.total_variance)
```

mbabadi (changes requested):

mbabadi (approved):
Nicely done! :-)

ordabayevy:
Thanks for reviewing @mbabadi !

PR #50, opened 2023-04-08
Python 3.10 & mypy

ordabayevy:


scvid/data/distributed_anndata.py
```diff
@@ -96,31 +96,29 @@ class DistributedAnnDataCollection(AnnCollection):
...
-        filenames: Union[Sequence[str], str],
-        limits: Optional[Sequence[int]] = None,
-        shard_size: Optional[int] = None,
-        last_shard_size: Optional[int] = None,
-        max_cache_size: Optional[int] = None,
+        filenames: Sequence[str] | str,
+        limits: Iterable[int] | None = None,
+        shard_size: int | None = None,
+        last_shard_size: int | None = None,
+        max_cache_size: int = 1,
```
ordabayevy, line 103: It is best to avoid `None` value here since `LRU(max_size=None)` will set to its own default `max_size=128`.

scvid/module/onepass_mean_var_std.py
```diff
@@ -36,9 +36,9 @@ def forward(self, x_ng: torch.Tensor) -> None:
         _, num_replicas = get_rank_and_num_replicas()
         if num_replicas > 1:
             x_ng = torch.cat(GatherLayer.apply(x_ng), dim=0)
-        self.x_sums += x_ng.sum(dim=0)
-        self.x_squared_sums += (x_ng**2).sum(dim=0)
-        self.x_size += x_ng.shape[0]
+        self.x_sums = self.x_sums + x_ng.sum(dim=0)
```
ordabayevy, line 39: Fixes runtime error (due to Python 3.10?): `RuntimeError: output with shape [] doesn't match the broadcast shape [5]`.

scvid/transforms/transforms.py
```diff
@@ -36,7 +34,7 @@ def __init__(
 
     def forward(self, x_ng: torch.Tensor) -> torch.Tensor:
         # Log1pNormalize
-        l_n1 = x_ng.sum(axis=-1, keepdim=True)
+        l_n1 = x_ng.sum(dim=-1, keepdim=True)
```
ordabayevy, line 37: Fixes some mypy error.

scvid/module/onepass_mean_var_std.py
mbabadi: Strange :)

scvid/transforms/transforms.py
mbabadi: Strange :)

mbabadi (approved):
I wonder how sustainable this is going to be in the long run 😅 My prior experience with mypy was that it got too much in the way. Also, I understand that certain classes from `Typing` are getting deprecated in 3.9+ and on should try to use `collections.abc` instead?

ordabayevy:
That's right, starting from [PEP585](https://peps.python.org/pep-0585/) `collections.abc` should be used.

> I wonder how sustainable this is going to be in the long run 😅

I wonder too. But getting mypy to pass for this repo wasn't as bad as I expected, maybe my expectations were low 😅.

PR #51, opened 2023-04-10
Typecheck `test` folder

ordabayevy:
copilot:summary


mbabadi (approved):
It was a very satisfying review :D like watching those satisfying Instagram videos that they cut into jellies, pool table that all of the balls go into the right holes, etc. 🤣

PR #61, opened 2023-07-01
Incremental PCA

ordabayevy:

PR #62, opened 2023-07-01
Embedding Writer

ordabayevy:


ordabayevy:
Blocked by #61

scvid/data/dadc_dataset.py
```diff
@@ -104,6 +104,7 @@ def __getitem__(self, idx: int | list[int] | slice) -> dict[str, np.ndarray]:
 
         data = {}
         data["X"] = X.toarray() if issparse(X) else X
+        data["obs_names"] = self.dadc[idx].obs_names.values.astype(int)
```
mbabadi, line 107: Could could comment on this?

ordabayevy: These are CAS IDs that are used during writing embeddings to files.

scvid/callbacks/prediction_writer.py
```diff
@@ -0,0 +1,73 @@
...
+        os.makedirs(output_dir, exist_ok=True)
+    df = pd.DataFrame(prediction.cpu())
+    df.insert(0, "db_ids", ids.cpu().numpy())
+    output_path = os.path.join(output_dir, f"batch_{postfix}.csv")
+    df.to_csv(output_path, header=False, index=False)
+
+
+class PredictionWriter(pl.callbacks.BasePredictionWriter):
+    """
+    Write prediction to a CSV file. The CSV file will have the same number of rows as the
```
mbabadi, line 38: predictions

scvid/data/dadc_dataset.py
mbabadi: Why convert to int? What if CAS IDs are non-int?

scvid/module/incremental_pca.py
```diff
@@ -121,3 +121,11 @@ def components_kg(self) -> torch.Tensor:
...
         """
         return self.V_kg
+
+    def predict(self, x_ng: torch.Tensor) -> torch.Tensor:
+        r"""
+        Embedding of the input data ``x_ng`` into the principal component space.
+        """
+        if self.transform is not None:
+            x_ng = self.transform(x_ng)
+        return x_ng @ self.V_kg.T
```
mbabadi, line 131: What about mean subtraction?

mbabadi (changes requested):

scvid/data/dadc_dataset.py
ordabayevy: I think it is a convention for CAS, right @fedorgrab ?

Returned data dictionary is collated and before that it is converted to a `torch.Tensor`. So it has to be a number. And I guess the only number type that makes sense for an ID is an `int` type.

scvid/module/incremental_pca.py
ordabayevy: Mean subtraction will just shift all the vectors by a constant vector, right? I thought in that case it shouldn't have any effect on vector similarity search.

scvid/module/base_module.py
```diff
@@ -21,6 +21,17 @@ def _get_fn_args_from_batch(
         Get forward method arguments from batch.
         """
 
+    def predict(self, x_ng: torch.Tensor) -> torch.Tensor | None:
```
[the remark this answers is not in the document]
ordabayevy: This `BaseModule` and subclasses that haven't defined the `predict` method return `None`.

test/test_distributed_anndata.py
```diff
@@ -149,6 +149,7 @@ def test_indexing(
         dat_view = dat[oidx, vidx]
         np.testing.assert_array_equal(adt_view.X, dat_view.X)
         np.testing.assert_array_equal(adt_view.var_names, dat_view.var_names)
+        np.testing.assert_array_equal(adt_view.obs_names, dat_view.obs_names)
```
ordabayevy, line 152: `obs_names` test

scvid/data/distributed_anndata.py
```diff
@@ -36,6 +38,38 @@ def lazy_getattr():
...
+
+            adata = self.adatas[i]
+            indices.append(adata.obs_names[oidx])
+
+        if len(indices) > 1:
+            concat_indices = pd.concat(
+                [pd.Series(idx) for idx in indices], ignore_index=True
+            )
+            obs_names = pd.Index(concat_indices)
+            obs_names = obs_names if self.reverse is None else obs_names[self.reverse]
```
ordabayevy, line 66: Reverse order if necessary. This ensures that `.X` and `.obs_names` have the same order.

scvid/module/base_module.py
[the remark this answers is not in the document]
fedorgrab: If this method's required for subclasses (which I think it is), I would better avoid making it optional to return tensor and require to return tensor.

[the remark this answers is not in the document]
mbabadi: How about `raise NotImplementedError` as a base implementation? That's the standard practice for abstract methods.

mbabadi (changes requested):

scvid/module/base_module.py
```diff
@@ -32,3 +33,21 @@ class BasePyroModule(pyro.nn.PyroModule, BaseModule, metaclass=PyroABCMeta):
...
+    def predict(self, x_ng: torch.Tensor) -> torch.Tensor:
+        """
+        Perform prediction on data tensor.
+
+        Args:
+            x_ng: Data tensor.
+
+        Returns:
+            Prediction tensor.
+        """
```
ordabayevy, line 53: Ended up using `PredictMixin` (@fedorgrab 👏 ), I think it is less code and more flexible rather than creating new abstract subclasses (have to duplicate subclasses for Pyro modules as well).

scvid/train/training_plan.py
```diff
@@ -70,6 +71,12 @@ def training_step(  # type: ignore[override]
             self.log("train_loss", loss)
         return loss
 
+    def forward(self, batch: dict[str, np.ndarray | torch.Tensor]) -> torch.Tensor:
+        """Forward pass of the model."""
+        assert isinstance(self.module, PredictMixin)
```
ordabayevy, line 76: Here check that `self.module` has a `.predict` method.

PR #65, opened 2023-08-04
`TDigest` module to calculate non-zero median

ordabayevy:


ordabayevy:
From PyTorch Lightning:

> Callbacks should capture NON-ESSENTIAL logic that is NOT required for your [lightning module](https://lightning.ai/docs/pytorch/stable/common/lightning_module.html) to run.
> 
> An overall Lightning system should have:
> 
> 1. Trainer for all engineering
> 2. LightningModule for all research code.
> 3. Callbacks for non-essential code.

I think based on this it is the best practice to have all model related code in one file/class (`torch.nn.Module`) - `forward` pass and `on_epoch_end`/`on_train_end` logic. Therefore I moved the merging logic from a callback to an `on_epoch_end` method of a torch module which is always (if it exists) called by `TrainingPlan` lightning module. This ensures that these model related hooks are bundled with the model and are called before any other callbacks (e.g. `ModuleCheckpoint`).

examples/tdigest.py
```diff
@@ -0,0 +1,51 @@
+# Copyright Contributors to the Cellarium project.
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""
+Example: One-pass calculation of feature median using t-digest statistics
+=========================================================================
+
+This example shows how to calculate non-zero median of log normalized feature count
```
mbabadi, line 8: I believe you're not log transforming, right?

examples/tdigest.py
```diff
@@ -0,0 +1,51 @@
...
+data in one pass [1].
+
+Example run::
+    python examples/tdigest.py fit \
```
mbabadi, line 12: In the future, this should also take the normalization config YAML.  We would bootstrap:  first, we run tdigest with a basic normalization config YAML that disables scaling by median and z-scoring (and provides no path to those checkpoints), but optionally enables log1p and/or library size normalization.  We then use the obtained checkpoint to make a new normalization config YAML.  We should check that tdigest checkpoint basic normalization config YAML is compatible with the new normalization config YAML.  For example, if we log1p before tdigest and making the checkpoint, the new normalization config YAML must also enable log1p, etc.

scvid/module/from_cli.py
```diff
@@ -129,3 +130,18 @@ def __init__(
...
+    """
+    Preset default values for the LightningCLI.
+
+    Args:
+        g_genes: Number of genes.
+        target_count: Target gene epxression count. Default: ``10_000``
+        eps: A value added to the denominator for numerical stability. Default: ``1e-6``
+    """
+
+    def __init__(self, g_genes, target_count: int = 10_000, eps: float = 1e-6) -> None:
```
mbabadi, line 145: In the future, these args should be replaced with a normalization config YAML (see my earlier comment).

mbabadi (changes requested):

examples/tdigest.py
ordabayevy: Good catch! Fixed it.

mbabadi (approved):

PR #70, opened 2023-08-18
`pyproject.toml`

ordabayevy:
- `setup.py`, `setup.cfg` -> `pyproject.toml`
- `flake8`, `isort` -> `ruff`
- `versioneer` (doesn't support pyproject.toml) -> `setuptools-git-versioning`


Makefile
```diff
@@ -7,16 +7,15 @@ uninstall: FORCE
 	pip uninstall scvi-distributed
 
 lint: FORCE
-	flake8
+	ruff check .
```
fedorgrab, line 10: I would recommend sticking to the same linting rules as CAS has. Which basically the same thing as you have, but the line length is 120. You can find the example here:
https://github.com/cellarium-ai/cellarium-cloud/blob/main/tox.ini#L47

We can review and discuss the preferred line length and may change it in CAS sometime later, but I think we should be consistent on this thing. Also, the default value in black is 88 which seems a little bit too short these days especially having the type annotations.

pyproject.toml
```diff
@@ -0,0 +1,64 @@
...
+classifiers = [
+  "Intended Audience :: Developers",
+  "Intended Audience :: Science/Research",
+  "License :: OSI Approved :: BSD License",
+  "Operating System :: OS Independent",
+  "Programming Language :: Python :: 3",
+]
+
+dependencies = [
+  "anndata",
```
fedorgrab, line 24: I recently had a problem with the CAS repo because I didn't specify versions for the packages. Just another suggestion is to provide specific versions or version ranges.

fedorgrab (approved):

PR #79, opened 2023-08-28
Deprecate single-consumer sampler non-iterable dataset classes

ordabayevy:
Also reduced the size of the test anndata files from `10_000` cells to `100` cells to prevent out of memory errors by github actions.


mbabadi (approved):
I always hate throwing out nice code, but I guess that's a good thing in the long run ... goodbye `DistributedAnnDataCollectionSingleConsumerSampler` :-)

PR #80, opened 2023-09-06
Publish to PyPI and TestPyPI

ordabayevy:


ordabayevy:
There is a user warning I get when I do the following import from the root directory of the repo:

```py
>>> from cellarium.ml import module
/home/yordabay/anaconda3/envs/cellarium/lib/python3.10/site-packages/lightning_utilities/core/imports.py:13: UserWarning: Module cellarium was already imported from None, but /mnt/disks/dev/repos/cellarium-ml is being added to sys.path
  import pkg_resources
```

It is due to the import of a deprecated `pkg_resources` package (https://setuptools.pypa.io/en/latest/pkg_resources.html) in lightning utilities and doesn't seem to affect the code. Also this doesn't happen from other directories.

mbabadi (approved):

ordabayevy:
Added a commit that fixes `configure_optimizer`s return type hint that was updated in https://github.com/Lightning-AI/lightning/pull/18463

PR #89, opened 2023-10-10
Move `examples` to `cellarium.ml.cli`

ordabayevy:
- Move `examples` to `cellarium.ml.cli`. This allows importing `main` functions and calling them. e.g. by the Vertex AI components to form pipelines
- Introduce `lightning_cli_factory` function to create new classes of `LightningCLI` on demand. This allows setting the default model and linking arguments
- Add `cellarium-ml` script in `pyproject.toml` to allow training models from the shell. Usage example: `cellarium-ml incremental_pca fit --config config.yaml`


mbabadi (approved):
Ah I like it! :-)

Issue #76, opened 2023-08-28
Refactor `transforms` module


ordabayevy:
- Break up `ZScoreLog1pNormalize` into `ZScore` and `Log1p` transforms (`NormalizeTotal` already implemented)
- Add additional metadata to transforms, e.g. total library size for `NormalizeTotal`
- Add validation schema to transforms

PR #90, opened 2023-10-13
Refactor `transforms`

ordabayevy:
Addresses #76


cellarium/ml/transforms/transforms.py
```diff
@@ -1,108 +1,251 @@
...
+        Args:
+            x_ng:
+                Gene counts.
+            feature_g:
+                The list of the variable names in the input data. If ``None``, no validation is performed.
+
+        Returns:
+            Gene counts divided by scale.
+        """
+        if self.feature_schema is not None and feature_g is not None:
```
mbabadi, line 55: Let's cache the validations:
```
@cache
def input_features_are_valid(self, feature_g: np.ndarray) -> bool:
    return np.array_equal(self.feature_schema, feature_g)

def forward(...):
    ...
    assert x_ng.shape[1] == len(feature_g), "The number of x_ng columns must match the feature_g length."
    assert self.input_features_are_valid(feature_g), "feature_g must match the feature schema".
    ...
```

... and I'm hoping `hash(feature_g)` is faster than `np.array_equal(self.feature_schema, feature_g)` ... right?!

cellarium/ml/transforms/transforms.py
```diff
@@ -1,108 +1,251 @@
...
+    @cache
+    def filter(self, feature_g: tuple) -> np.ndarray[Any, np.dtype[np.bool_]]:
+        """
+        Args:
+            feature_g: The list of the variable names in the input data.
+
+        Returns:
+            A boolean mask of the features to filter by.
+        """
+        return np.isin(feature_g, self.filter_list)
```
mbabadi, line 94: assert than the mask has at least one non-zero element.

cellarium/ml/transforms/transforms.py
```diff
@@ -1,108 +1,251 @@
...
+            The variable names schema for the input data validation.
+            If ``None``, no validation is performed.
     """
 
-    def __init__(
-        self,
-        target_count: int = 10_000,
-        eps: float = 1e-6,
-    ):
+    def __init__(self, scale_g: torch.Tensor, eps: float = 1e-6, feature_schema: ArrayLike | None = None) -> None:
```
mbabadi, line 31: Let's not allow the case where feature_schema is None. Instead, let's use caching to make validations fast (see below).

cellarium/ml/transforms/transforms.py
```diff
@@ -1,108 +1,251 @@
...
+
+        y_{ng} = x_{ng}[:, \\mathrm{mask}_g]
+
+    Args:
+        filter_list: A list of features to filter by.
+    """
+
+    def __init__(self, filter_list: ArrayLike) -> None:
+        super().__init__()
+        self.filter_list = np.array(filter_list)
```
mbabadi, line 83: assert that filter_list is non-empty.

cellarium/ml/transforms/transforms.py
```diff
@@ -1,108 +1,251 @@
...
-    def __init__(self, scale_g: torch.Tensor, eps: float = 1e-6):
+    def __init__(
+        self,
+        target_count: int = 10_000,
+        eps: float = 1e-6,
+    ) -> None:
         super().__init__()
-        self.scale_g: torch.Tensor
-        self.register_buffer("scale_g", scale_g)
+        self.target_count = target_count
```
mbabadi, line 165: assert that target_count is positive.

cellarium/ml/transforms/transforms.py
```diff
@@ -1,108 +1,251 @@
...
         self.eps = eps
```
mbabadi, line 166: assert that eps is positive here and elsewhere.

cellarium/ml/transforms/transforms.py
mbabadi: write utility functions `assert_positive(Number) -> Number` and `assert_nonnegative(Number) -> Number`.

cellarium/ml/transforms/transforms.py
ordabayevy: Another consideration is that `np.ndarray` is not hashable and first needs to be converted to `tuple` (hashable). Then the question is is `hash(tuple(feature_g.tolist()))` is faster than `np.array_equal(self.feature_schema, feature_g)`?

cellarium/ml/transforms/transforms.py
ordabayevy: I will do the same for `ZScore` then?

cellarium/ml/transforms/transforms.py
ordabayevy: Maybe we can we `id(feature_g)` as a hash? Example of that in funsor: https://github.com/pyro-ppl/funsor/blob/c96e70c89aaa0bda8dc132ae1c3498a681f618b8/funsor/interpretations.py#L66

ordabayevy: Some tests:

```py
In [4]: import numpy as np

In [5]: x = np.array([f"gene_{i}" for i in range(30000)])

In [6]: x
Out[6]: 
array(['gene_0', 'gene_1', 'gene_2', ..., 'gene_29997', 'gene_29998',
       'gene_29999'], dtype='<U10')

In [7]: %timeit x.tolist()
1.23 ms ± 6.69 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [8]: %timeit tuple(x.tolist())
1.46 ms ± 10.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [9]: %timeit hash(tuple(x.tolist()))
2.14 ms ± 24.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [10]: %timeit np.array_equal(x, x)
325 µs ± 2.87 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

cellarium/ml/transforms/transforms.py
```diff
@@ -1,108 +1,261 @@
...
+            raise ValueError(f"`filter_list` must not be empty. Got {self.filter_list}")
+
+    @cache
+    def filter(self, feature_g: tuple) -> np.ndarray[Any, np.dtype[np.bool_]]:
+        """
+        Args:
+            feature_g: The list of the variable names in the input data.
+
+        Returns:
+            A boolean mask of the features to filter by.
```
mbabadi, line 99: tuple(ndarray.tolist()) is a bottleneck. If feature_g (ndarray) is hashable, let's not transform it here and let filter do it.

cellarium/ml/module/from_cli.py
```diff
@@ -252,6 +253,7 @@ def __init__(
                 ),
                 DivideByScale(
                     scale_g=tdigest.median_g,
+                    feature_schema=feature_schema,
```
mbabadi, line 256: We should support the case where tdigest also has Log1p.

mbabadi (changes requested):
Nice work!  Let's iterate on this one a bit more.

mbabadi (approved):

Issue #92, opened 2023-10-17
Clean up the submodule structure


ordabayevy:
- [x] Try to follow PyTorch Lightning's submodule structure: https://github.com/Lightning-AI/lightning/tree/master/src/lightning/pytorch
- [x] Rename `TrainingPlan` to `CellariumModule` and move to the `cellarium.ml.core` submodule and remove `cellarium.ml.train`
- [x] Rename `cellarium.ml.module` to `cellarium.ml.models` submodule
- [x] Move `DistributedAnnDataCollectionDataModule` to `cellarium.ml.core` submodule
- [x] Move `read.py` to `fileio.py`
- [x] Rename `cellarium.ml.optim` to `cellarium.ml.lr_schedulers`
- [x] Move `test` to `tests`

PR #96, opened 2023-10-22
Refactor submodules

ordabayevy:
Addresses #92


tests/common.py
```diff
@@ -6,7 +6,7 @@
 import torch
 
 
-class TestDataset(torch.utils.data.Dataset):
+class BoringDataset(torch.utils.data.Dataset):
```
mbabadi, line 9: Nice :D

mbabadi (approved):
Very nice reorg!  I am approving this PR w/ some comments for a future feature:

I think the next logical refactoring work is to automate the generation of all of the `XFromCLI` classes + automatic registration in the `cli.py`. It is more natural to move the instructions needed to make `XFromCLI` to the model implementations under `/models` instead of having them all under `from_cli.py`. We could also rename `XFromCLI` to `XStandAlone` (or `XRunnable` or `XTool`). Everything else that is needed for lightning_cli_factory could also be included in the same class somewhere (e.g. default arguments, link_arguments, etc) such that `/cli.py` would become completely generic, e.g. just go over all of the models, chose the ones that are flagged for registration to CLI via decorator, and register all of them.

tests/common.py
ordabayevy: `pytest` attempts to parse it as a test everything that starts with `Test`. It understands that it is not a test but still prints a message about it.

PR #104, opened 2023-12-01
crick

ordabayevy:
- Add `crick` to dependency
- Save `tdigest` files with `trainer.strategy.checkpoint_io.save_checkpoint` which supports saving files to `gs://`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,3 +84,4 @@ line-length = 120
 [tool.mypy]
 ignore_missing_imports = true
 explicit_package_bases = true
+check_untyped_defs = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ dependencies = [
   "anndata",
   "boltons",
   "braceexpand",
-  "crick",
+  "crick>=0.0.4",
   "google-cloud-storage",
   "jsonargparse[signatures]",
   "lightning>=2.0.9",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,7 @@ dependencies = [
   "anndata",
   "boltons",
   "braceexpand",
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
   "anndata",
   "boltons",
   "braceexpand",
+  "crick",
```
fedorgrab, line 27: Could you please specify a version you're using? Otherwise I can get into my problem with crick again after I update cellarium-ml to the latest version..

tests/test_tdigest.py
```diff
@@ -110,8 +109,10 @@ def test_tdigest_multi_device(
         np.testing.assert_allclose(expected_median, actual_median, rtol=0.01)
 
 
-@requires_crick
-def test_load_from_checkpoint_multi_device(tmp_path: Path):
+def test_load_from_checkpoint_multi_device():
+    # using pytest tmp_path fixture creates a directory per rank
+    # so we need to use a fixed shared directory
+    tmp_path = Path("/tmp/test_load_from_checkpoint_multi_device")
```
fedorgrab, line 115: I am not entirely sure since I am still new to using tempfile and tempdirs, but perhaps this approach would be better

https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

ordabayevy: The problem with `tempfile` is that it creates a random directory each time it is called. When `devices=2` pytorch lightning calls the `test_load_from_checkpoint_multi_device` script again for the second device and `tempfile` will create a different random directory.

But I like the idea of cleaning up so I added `shututil.rmtree(tmp_path)` to clean up the directory.

fedorgrab (approved):

PR #106, opened 2023-12-14
ZScore register buffer fix, add to .gitignore

bricewang:
Register buffers in ZScore to synchronize mean/std tensors with the torch module

Add .ipynb_checkpoints to .gitignore

### DIFF
--- a/cellarium/ml/transforms/z_score.py
+++ b/cellarium/ml/transforms/z_score.py
@@ -41,8 +41,10 @@ class ZScore(nn.Module):
         eps: float = 1e-6,
     ) -> None:
         super().__init__()
-        self.mean_g = mean_g
-        self.std_g = std_g
+        self.mean_g: torch.Tensor
+        self.std_g: torch.Tensor
+        self.register_buffer("mean_g", torch.as_tensor(mean_g))
+        self.register_buffer("std_g", torch.as_tensor(std_g))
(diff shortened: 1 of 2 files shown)


cellarium/ml/transforms/z_score.py
```diff
@@ -41,8 +41,10 @@ def __init__(
         eps: float = 1e-6,
     ) -> None:
         super().__init__()
-        self.mean_g = mean_g
-        self.std_g = std_g
+        self.mean_g: torch.Tensor
+        self.std_g: torch.Tensor
+        self.register_buffer("mean_g", mean_g)
```
ordabayevy, line 46: Can you convert `mean_g` and `std_g` to torch.Tensor with `torch.as_tensor(mean_g)` and  `torch.as_tensor(std_g)`? `register_buffer` expects `torch.Tensor` but `mean_g` and `std_g` can also be floats.

ordabayevy (changes requested):
Looks great! Just a comment about types so that `mypy` would be happy.

cellarium/ml/transforms/z_score.py
bricewang: Got it, thanks! Much better than my idea of checking for float and using `torch.Tensor([mean_g])`

ordabayevy (approved):

PR #109, opened 2024-01-11
Fix the docs

ordabayevy:
- Remove `from __future__ import annotations` (type hints are rendered better without it)
- Fix the stacking of `py:property` on a single line (https://github.com/readthedocs/sphinx_rtd_theme/issues/1301#issuecomment-1876120817)
- Use `sphinx.ext.autodoc` for displaying type hints instead of `sphinx_autodoc_typehints`
- Remove `__weakref__` method from the docs

Before:
Screenshot 2024-01-11 at 10 33 23 AM

After:
Screenshot 2024-01-11 at 10 39 30 AM

### DIFF
--- a/cellarium/ml/core/module.py
+++ b/cellarium/ml/core/module.py
@@ -1,8 +1,6 @@
 # Copyright Contributors to the Cellarium project.
 # SPDX-License-Identifier: BSD-3-Clause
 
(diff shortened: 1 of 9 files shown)


fedorgrab (approved):
:+1:

PR #120, opened 2024-01-29
`compute_var_names_g`

ordabayevy:
- `compute_var_names_g` computes `var_names_g` by applying transforms to batch (of single cell) of data under fake mode (this ensures that no actual torch computation is performed)
- rename `get_n_categories` to `compute_n_categories`
- create `compute_n_obs` that is used instead of `n_obs` attribute in `CellariumAnnDataDataModule`
- use `LinkArguments` dataclass to pass args to `link_arguments`

### DIFF
--- a/cellarium/ml/core/pipeline.py
+++ b/cellarium/ml/core/pipeline.py
@@ -34,13 +34,15 @@ class CellariumPipeline(torch.nn.ModuleList):
     """
 
     def forward(self, batch: dict[str, np.ndarray | torch.Tensor]) -> dict[str, torch.Tensor | np.ndarray]:
+        from cellarium.ml import CellariumModule
+
         for module in self:
-            # get the module input keys
-            ann = module.forward.__annotations__
-            if "batch" in ann and len(ann) == 2:
+            if isinstance(module, CellariumModule):
                 # in case module is a CellariumModule, e.g. PCA checkpoint is used as a transform
                 batch |= module(batch)
             else:
+                # get the module input keys
+                ann = module.forward.__annotations__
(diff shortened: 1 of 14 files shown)


cellarium/ml/core/pipeline.py
```diff
@@ -37,11 +37,15 @@ def forward(self, batch: dict[str, np.ndarray | torch.Tensor]) -> dict[str, torc
         for module in self:
             # get the module input keys
             ann = module.forward.__annotations__
-            input_keys = {key for key in ann if key != "return" and key in batch}
-            # allow all keys to be passed to the module
-            if "kwargs" in ann:
-                input_keys |= batch.keys()
-            batch |= module(**{key: batch[key] for key in input_keys})
+            if "batch" in ann and len(ann) == 2:
```
mbabadi, line 40: Why not simply `isinstance(module, CelariumModule)`?

mbabadi (approved):
Good work!

General comment: I find it a bit clunky that we have to specify the model name as the first CLI arg just to invoke the appropriate argument linking, whereas the YAML file fully specifies the model. In that sense, specifying the model twice is superfluous.

Ideally, we'd want:
```
cellarium-ml fit --config my_incremental_pca_train_config.yaml
```
rather than:
```
cellarium-ml incremental-pca fit --config my_incremental_pca_train_config.yaml
```

We should figure out a workaround at a later time :) Do you agree?

ordabayevy:
I agree that it is a bit annoying having to have specify the model twice but don't have any immediate solution in mind.

PR #125, opened 2024-02-06
Update CellariumAnnDataDataModule

ordabayevy:
- Add a warning filter for loading anndata files
- Save hyper-parameters by using `save_hyperparameters`
  - `save_hyperparameters` pulls init args from the `locals()`. Therefore, we can **uninitialize** `torch.nn.Modules` before calling `save_hyperparameters`.
  - Initialize `torch.nn.Module`s in the `__init__` method if they were provided as `class_path` and `init_args`.
  - Remove `config` arg from the `CellariumModule` that was used to store hyper-parameters
- Refactor `CellariumAnnDataDataModule` where both `DistributedAnnDataCollection` and `AnnData` can be passed
  - Remove a duplicated example in the docstring
- Remove default lr (0.001) and default optimizer (Adam). For models like `IncrementalPCA` or `OnePassMeanVarStd` optimizer is not needed.


cellarium/ml/core/module.py
```diff
@@ -59,33 +52,27 @@ def __init__(
...
         scheduler_fn: type[torch.optim.lr_scheduler.LRScheduler] | None = None,
         scheduler_kwargs: dict[str, Any] | None = None,
-        default_lr: float = 1e-3,
-        config: dict[str, Any] | Namespace | None = None,
     ) -> None:
         super().__init__()
+
+        _transforms, _model = transforms, model
+
+        transforms = [uninitialize_object(transform) for transform in _transforms] if _transforms is not None else None
```
mbabadi, line 60: Can you explain this to me?

mbabadi: Perhaps also write two lines of comment about why you're uninitializing and reinitializing?

mbabadi (changes requested):
Good work, nothing strikes my eyes :) Small request for comment on a magical-looking part of the code.

cellarium/ml/core/module.py
ordabayevy: Added a comment above explaining the logic behind un-initializing and initializing.

mbabadi: model and transforms are not attributes of self (yet)-- does uninitializing them have any effect on the behavior of `self.save_hyperparameters`?

cellarium/ml/core/datamodule.py
```diff
@@ -80,6 +80,15 @@ def __init__(
...
+        # anndata objects in the model checkpoint.
+        # Instead, we save the class name and the init args which then can be used to
+        # re-initialize the distributed anndata.
+        # In order to achieve this, we temporarily re-assign `dadc` to its un-initialized state
+        # and then call `save_hyperparameters` which will save these values as hparams.
+        # Then, we re-assign `dadc` back to its initialized state.
+        # `initialize_object` handles the case when the object was passed as a dictionary of class path and init args.
         _dadc = dadc
         dadc = uninitialize_object(_dadc)
         self.save_hyperparameters(logger=False)
```
mbabadi, line 94: can this be done more neatly with a context manager construct? also, `dadc` not not an attrib of self yet-- how would uninitializing it on line 93 have an effect on `self.save_hyperparameters` call on line 94?

ordabayevy: Having a context manager is good idea, let me think about it more.
`save_hyperparameters` looks up from the variable from the `locals()` using a argument name from the function signature.

ordabayevy: Another approach is to have `model` and `transforms` arg of `CellariumModule` as dicts of class name and init args and delegate their initialization to `CellariumModule` instead of `LightningCLI`. In fact, there is a `configure_model` method of `LightningModule` that is used to initialize large models with FSDP and DeepSpeed strategies (https://lightning.ai/docs/pytorch/stable/advanced/model_init.html#model-parallel-training-fsdp-and-deepspeed). We can just always use it to initialize all our models.

mbabadi (approved):
Good stuff all around!  I like initialization in meta device context + configure_model.

PR #127, opened 2024-02-26
Fix concatenating anndata files

### LINKED ISSUE #112: Logistic Regression train doesn't work with multiple GPUs
When I start a training script with multiple GPUs, I get:

```
File "/home/user/micromamba/lib/python3.10/site-packages/cellarium/ml/utilities/data.py", line 194, in categories_to_codes
AttributeError: 'numpy.ndarray' object has no attribute 'cat'
```

ordabayevy:
Resolves #112

### DIFF
--- a/cellarium/ml/data/distributed_anndata.py
+++ b/cellarium/ml/data/distributed_anndata.py
@@ -5,9 +5,10 @@ import gc
 from collections.abc import Iterable, Sequence
 from contextlib import contextmanager
 
+import numpy as np
 import pandas as pd
-from anndata import AnnData
-from anndata._core.index import Index, _normalize_indices
+from anndata import AnnData, concat
+from anndata._core.index import Index, Index1D, _normalize_indices
 from anndata.experimental.multi_files._anncollection import (
     AnnCollection,
     AnnCollectionView,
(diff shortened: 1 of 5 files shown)


cellarium/ml/data/distributed_anndata.py
```diff
@@ -212,43 +213,44 @@ def __init__(
...
-        self.materialize(adatas_indices)
-
-        return DistributedAnnDataCollectionView(self, self.convert, resolved_idx)
+        adatas_oidx, oidx, vidx, reverse = self._resolve_idx(oidx, vidx)
+        adatas = self.materialize(adatas_oidx, vidx)
+        adata = concat(adatas, merge="same")
+        adata = adata if reverse is None else adata[reverse]
+        # make sure that categorical dtypes are preserved
+        adata.obs = adata.obs.astype(self.schema.attr_values["obs"].dtypes)
+        return adata
```
ordabayevy, line 229: Now this returns a single gathered anndata object instead of `DistributedAnnDataCollectionView` thus bypassing the logic from `AnnCollection`. `DistributedAnnDataCollectionView` is only useful if we slice lazily multiple times which we never do.

Another difference here is that now we gather and concatenate entire anndata objects at once. With `DistributedAnnDataCollectionView` gathering/concatenating happens for each attribute that is accessed (`X`, `obs`, etc) which in edge cases can lead to loading the same anndata file multiple times when multiple attributes are accessed.

tests/test_distributed_anndata.py
```diff
@@ -239,6 +247,7 @@ def dadc(adatas_path: Path):
     [
         ("X", None, None),
         ("obs", "A", None),
+        ("obs", "C", categories_to_codes),
```
ordabayevy, line 250: This test exercises the case from the logistic regression bug.

mbabadi (approved):
Looks good! :)

PR #129, opened 2024-03-07
CellariumGPT

ordabayevy:


cellarium/ml/models/cellarium_gpt.py
```diff
@@ -0,0 +1,745 @@
...
+        #   > the weights of residual layers at initialization by a factor of 1/√N where N is the # of residual layers.
+        #   >   -- GPT-2 :: https://openai.com/blog/better-language-models/
+        #
+        # Reference (Megatron-LM): https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/model/gpt_model.py
+        for name, p in module.named_parameters():
+            if name == "dense2.weight" or name == "Wo.weight":
+                # Special Scaled Initialization --> There are 2 Layer Norms per Transformer Block
+                p.data.normal_(mean=0.0, std=(self.initializer_range / math.sqrt(2 * self.gpt_model.n_blocks)))
+
+    def tokenize(
```
mbabadi, line 584: documentation

mbabadi: Can you also separately take `obs_total_mrna_umis_n` (to be concatenated to `values_nc`) and `out_total_mrna_umis_n` to generate a token specifying the library size we need the _readout_ to be at? This construct can be used for model training with downsampling. We could use Brice's logic to `Duplicate` the counts, downsample the first copy to use for observation, and use the second copy for readout.

Another thought is to have multiple `tokenize` methods. For example:
-- `generate_observation_tokens`
-- `generate_output_tokens`
-- `generate_register_tokens` (later)
-- `generate_metadata_tokens` (later)

In this construct, `generate_observation_tokens` takes `x_ng`, `obs_total_mrna_umis_n` as input and generates a bunch of tokens.  `generate_output_tokens`, for now, just takes `out_total_mrna_umis_n` and generates a single token.

All of the tokens are then concatenated and given to an embedding layer. The tokens should also be equipped with metadata to specify how they should be embedded ...

cellarium/ml/models/cellarium_gpt.py
```diff
@@ -0,0 +1,745 @@
...
+
+        # log_theta_mu_eps = torch.log(self.theta + self.mu + self.eps)
+        # return (
+        #     self.theta * (torch.log(self.theta + self.eps) - log_theta_mu_eps)
+        #     + value * (torch.log(self.mu + self.eps) - log_theta_mu_eps)
+        #     + torch.lgamma(value + self.theta)
+        #     - torch.lgamma(self.theta)
+        #     - torch.lgamma(value + 1)
+        # )
+        delta = torch.where(
```
mbabadi, line 69: Add documentation:

"For large "theta", we can use Stirling's asymptotic approximation (see https://en.wikipedia.org/wiki/Gamma_function#Log-gamma_function), which is numerically more stable than PyTorch's implementation of lgamma."

Actually, the condition value / theta < 1e-2 may not be needed. Let's make an issue for me to investiagte.

cellarium/ml/cli.py
```diff
@@ -318,6 +318,23 @@ def add_arguments_to_parser(self, parser: LightningArgumentParser) -> None:
     return NewLightningCLI
 
 
+@register_model
+def cellarium_gpt(args: ArgsType = None) -> None:
+    r"""
+    CLI to run the :class:`cellarium.ml.models.CellariumGPT` model.
```
mbabadi, line 324: Add example CLI command.

mbabadi (changes requested):
Some small changes here and there.

cellarium/ml/models/cellarium_gpt.py
mbabadi: Make an issue (#131).

cellarium/ml/models/cellarium_gpt.py
```diff
@@ -0,0 +1,774 @@
...
+        self.output_mult = output_mult
+
+    def forward(self, hidden_state_ncd) -> tuple[torch.Tensor, torch.Tensor]:
+        output = self.dense(hidden_state_ncd) * self.output_mult
+        mu_nc = output[..., 0].exp()
+        theta_nc = output[..., 1].exp()
+        return mu_nc, theta_nc
+
+
+class CellariumGPT(CellariumModel):
```
sjfleming, line 479: Can you make it a `PredictMixin` and implement `predict()` ?

sjfleming: Although I'm thinking that, analogous to Geneformer, `predict()` would return the gene embedding vectors.  But I'm not sure that exactly makes sense here... you might also imagine that "predict" would return the negative binomial distributions.

My interest was in having a common sort of interface I could use to extract gene embeddings, whether or not the model was a Geneformer model or a CellariumGPT model.  I was previously using `.predict()` to do this from Geneformer, so I thought it would be nice if CellariumGPT worked the same way.

cellarium/ml/models/cellarium_gpt.py
```diff
@@ -0,0 +1,774 @@
...
+        labels_mask_nc = torch.arange(self.n_context, device=device)[None, :] >= prefix_len_n[:, None]
+
+        ids_nc, values_nc = self.tokenize(x_ng, total_mrna_umis_n, self.n_context, shuffle=True)
+        hidden_state_ncd = self.gpt_model(ids_nc, values_nc, labels_mask_nc, prefix_len_n, "block_diagonal")
+        mu_nc, theta_nc = self.nb_head(hidden_state_ncd)
+
+        sample_weights_nc = 1 / labels_mask_nc.sum(dim=1, keepdim=True).expand(-1, self.n_context)
+        nb_dist = NegativeBinomial(mu=mu_nc[labels_mask_nc], theta=theta_nc[labels_mask_nc])
+        log_prob = nb_dist.log_prob(values_nc[labels_mask_nc])
+        sample_weights = sample_weights_nc[labels_mask_nc]
```
sjfleming, line 632: Is this supposed to be like this?
```python
sample_weights_nc = sample_weights_nc[labels_mask_nc]
```

cellarium/ml/models/cellarium_gpt.py
```diff
@@ -0,0 +1,774 @@
...
+        # prefix includes the total_mrna_umis and a random subset of genes
+        # the length of the prefix is sampled uniformly from 1 to n_context - 1 (inclusive)
+        prefix_len_n = torch.randint(1, self.n_context, (n,), device=device)
+        labels_mask_nc = torch.arange(self.n_context, device=device)[None, :] >= prefix_len_n[:, None]
+
+        ids_nc, values_nc = self.tokenize(x_ng, total_mrna_umis_n, self.n_context, shuffle=True)
+        hidden_state_ncd = self.gpt_model(ids_nc, values_nc, labels_mask_nc, prefix_len_n, "block_diagonal")
+        mu_nc, theta_nc = self.nb_head(hidden_state_ncd)
+
+        sample_weights_nc = 1 / labels_mask_nc.sum(dim=1, keepdim=True).expand(-1, self.n_context)
```
sjfleming, line 629: Is it true that with no masking for cell 0, i.e. `labels_mask_nc[0, :].sum(1) == 0`, then `sample_weights_nc[0, :] == inf` ?  Is this desired?

sjfleming: I suppose there is always some masking during training.  I was thinking about prediction where maybe I would not mask anything, and I was seeing `inf` values (though I would not use `sample_weights` anyway).

sjfleming: Oh I see you've enforced this anyway by saying the max range of the prefix size is `n_context - 1`, looks like you'll never have that 1 / 0 situation.

cellarium/ml/models/cellarium_gpt.py
sjfleming: See #152 if you're interested in this idea

cellarium/ml/models/cellarium_gpt.py
ordabayevy: That's right. `sample_weights_nc` is only needed during training to calculate the loss.

cellarium/ml/models/cellarium_gpt.py
ordabayevy: I've been locally implementing different versions of `predict` method depending on what I wanted to analyze. Since I was changing it a lot decided not to add it here. But it should be something similar to Geneformer predict that returns a dictionary and we can have boolean flags that can be used to control what it returns.

PR #150, opened 2024-03-21
Refactor the negative binomial code

### LINKED ISSUE #148: scvi - refactor the NB distribution
For now we have copied and pasted way too much code from scvi-tools when it comes to the negative binomial distribution.

Get rid of a lot of that code.

Only keep what is necessary.

Also, @mbabadi has pointed out 
> This is not robust when theta is large, see cellarium-gpt branch.

Let's leverage the NB distribution that's already been written by @ordabayevy on the cellarium-gpt branch and get rid of all the scvi-tools copy and paste.

sjfleming:
Closes #148 

Use the code Yerdos was using in the `cellarium-gpt` branch here

https://github.com/cellarium-ai/cellarium-ml/blob/fc8a2d079222daee6b7222f952e43c44e66ca931/cellarium/ml/models/cellarium_gpt.py#L27-L81

This should make it easy to have the two models use the same code.


sjfleming:
I added a helper mixin to log learning rate here as well

sjfleming:
I modified the scvi_pbmc_config.yaml
to get it to run well for me.  It would not run as it was.

sjfleming:
I added a demo scvi notebook that shows the UMAPs before and after.
Running this notebook could be a way to benchmark future code changes.

This NB distribution produces these results, which I think are probably the same as before:

## Before scVI
image

## After scVI
image

ImXman (commented):
Reviewed

ImXman (commented):
No comments

Issue #147, opened 2024-03-20
scvi - implement `predict()`


sjfleming:
Implement the `predict()` method of the scvi model.  This is just creating the embeddings given raw input data.

sjfleming:
Very likely it can just be
```python
        return self.inference(
            x=x_ng,
            batch_index=batch_index_n,
            cont_covs=cont_covs_nc,
            cat_covs=cat_covs_nd,
            n_samples=1,
        )
```

PR #151, opened 2024-03-22
Implement reset_parameters() for scvi

### LINKED ISSUE #146: scvi - implement `reset_parameters()`
And also figure out what it's doing now, when we just `pass`.

ImXman:
1) Add reset_parameters to scvi.py. It specifies how to initialize model weights
2) An updated config yaml to run scvi is added to example_config/

Closes #146

### DIFF
--- a/cellarium/ml/models/scvi.py
+++ b/cellarium/ml/models/scvi.py
@@ -17,6 +17,14 @@ from cellarium.ml.utilities.testing import (
     assert_columns_and_array_lengths_equal,
 )
 
+def weights_init(m):
+    classname = m.__class__.__name__
+    if classname.find('BatchNorm') != -1:
+        torch.nn.init.normal_(m.weight, 1.0, 0.02)
+        torch.nn.init.zeros_(m.bias)
+    elif classname.find('Linear') != -1:
+        torch.nn.init.xavier_normal_(m.weight)
+        torch.nn.init.zeros_(m.bias)
 
 def one_hot(index: torch.Tensor, n_cat: int) -> torch.Tensor:
     """One hot a tensor of categories."""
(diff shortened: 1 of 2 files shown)


sjfleming:
Also addresses #147

cellarium/ml/models/scvi.py
```diff
@@ -256,18 +264,14 @@ def __init__(
...
-        # rng.manual_seed(self.seed)
-        # if isinstance(self.mean_g, torch.nn.Parameter):
-        #     self.mean_g.data.zero_()
-        # self.W_kg.data.normal_(0, self.W_init_scale, generator=rng)
-        # self.sigma_unconstrained.data.fill_(
-        #     _unconstrain(
-        #       torch.as_tensor(self.sigma_init_scale), constraints.positive))
+        self.reset_parameters()
+
+        def reset_parameters(self) -> None:
```
sjfleming, line 269: This block of code is indented too much.  Four spaces need to be removed.

cellarium/ml/models/scvi.py
```diff
@@ -17,6 +17,14 @@
     assert_columns_and_array_lengths_equal,
 )
 
+def weights_init(m):
```
sjfleming, line 20: Is this function from somewhere else in the codebase?  If so, maybe we should consider refactoring so there is no copy-and-paste.

For now this is okay.

cellarium/ml/models/scvi.py
```diff
@@ -484,29 +488,16 @@ def predict(
         self,
         x_ng: torch.Tensor,
         var_names_g: np.ndarray,
-        transform_batch_n: torch.Tensor,
-    ) -> dict[str, np.ndarray | torch.Tensor]:
-        """
-        Embed the input data ``x_ng`` into the scVI latent space.
```
sjfleming, line 490: Please add documentation back in for `inference()`

sjfleming (changes requested):
Thanks Yang!  I think that the `reset_parameters()` function is overly indented, which will probably cause problems.  Also can you add back documentation for the `predict()` function?

PR #152, opened 2024-03-24
Add `predict()` to Cellarium GPT

sjfleming:
@ordabayevy this is only if you think it's a good idea.  Feel free to reject the idea.

I was trying to come up with a way to jam data through both Geneformer and Cellarium GPT in a somewhat-uniform way to get embeddings and maybe mu predictions from Cellarium GPT.  I was thinking `predict()` might be a good way to do that.


sjfleming:
Closing as this is defunct

Issue #113, opened 2024-01-22
Fix floating point overflow issues in `OnePassMeanVarStd`


ordabayevy:
At some point, this implementation _will_ bite us in the ass due to float overflows, and we won't know when :)

Here's a little test that I did:
```
n_trials = 50_000_000
mean = 151.
std = 0.1
dtype = torch.float64
var_actual = std ** 2
x = mean + std * torch.randn(n_trials, dtype=dtype)
var_emp = torch.sum(x ** 2) / n_trials - (torch.sum(x) / n_trials) ** 2
print(var_emp / var_actual)
```

It works well for float64, you get 1 +/- 0.0001. For float32, you get 1 +- 0.2 (!).  For float16, you get nan. If we're using float32, we're already getting garbage. I strongly suggest making the dtype of the running sums to float64 for now. Let's also make an issue to replace this implementation with a more numerically robust implementation (low priority).

_Originally posted by @mbabadi in https://github.com/cellarium-ai/cellarium-ml/pull/108#discussion_r1456251367_

PR #154, opened 2024-03-27
Add shifted data algorithm for variance computation to `OnePassMeanVarStd`

ordabayevy:
Addresses #113


cellarium/ml/models/onepass_mean_var_std.py
```diff
@@ -64,9 +70,28 @@ def forward(self, x_ng: torch.Tensor, var_names_g: np.ndarray) -> dict[str, torc
...
 
-        self.x_sums = self.x_sums + x_ng.sum(dim=0)
-        self.x_squared_sums = self.x_squared_sums + (x_ng**2).sum(dim=0)
-        self.x_size = self.x_size + x_ng.shape[0]
+        if self.algorithm == "naive":
+            self.x_sums = self.x_sums + x_ng.sum(dim=0)
+            self.x_squared_sums = self.x_squared_sums + (x_ng**2).sum(dim=0)
+            self.x_size = self.x_size + x_ng.shape[0]
+        elif self.algorithm == "shifted_data":
+            if self.x_shift is None:
```
mbabadi, line 78: So... you get shift from the very first minibatch? I think it is an improvement over naive, but doesn't really close the issue ;-) we should implement Welford's online algorithm for that: https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance

mbabadi (approved):

cellarium/ml/models/onepass_mean_var_std.py
ordabayevy: I have worked on that initially but then it turned out to be that it is either unstable or hard to implement merging results from multiple GPUs. Can't remember which one it was 🙃

PR #155, opened 2024-03-27
Refactor common model code

### LINKED ISSUE #140: scvi - refactor common
Regarding `Encoder` and `DecoderSCVI`, @mbabadi said

> As a matter of principle, things that belong to common must be common :D DecoderSCVI is not common, because it is only used by SCVI. Either refactor it to something more general-purpose, or refactor out to scvi.py.

Transition from scvi-tools' code organization to an organization that makes more sense in cellarium-ml.

sjfleming:
Closes #140


sjfleming:
Everything was tested and seems to still work the same

image

sjfleming:
Tested after the last commit and it still works

Issue #145, opened 2024-03-20
scvi - promote batch bias vectors to latent variables


sjfleming:
We would like to have the option to promote batch bias vectors to latent variables.

This amounts to doing two things during inference:

- augment the loss function with a new KL divergence
- learn the distribution of each $d$-dimensional batch $s$ bias vector (for example, $[\mu_{sd}, \sigma_{sd}]$), and sample from that distribution when computing the reconstruction loss according to whatever variational posterior we choose: $W_{sd} \sim \mathcal{N} (\mu_{sd}, \sigma_{sd})$

<img width="413" alt="image" src="https://github.com/cellarium-ai/cellarium-ml/assets/10214815/7f0e41ec-5038-4b3f-b8ce-1eac01fbf3f6">

sjfleming:
The encoder and decoder also need to have additional flexibility added... they need to accommodate this different parameterization.

Don't just have the one-hot concatenation as the only option... allow a batch bias vector to be added.

sjfleming:
Probably want to make a subclass of `torch.nn.Linear` in order to accommodate a learnable bias vector.

sjfleming:
We ultimately want to learn not the bias vectors per batch by themselves, but their _distributions_.  So maybe like a mean and a variance.

ImXman:
I am thinking of a forward function that takes input (x, b). torch.nn.Linear will take x as input to get out1, and an embedding layer takes b to project batch label to out2. The final output would be out=out1+out2.

ImXman:
I looked up the implementation in scvi-tools. The way they implemented batch as embedding is concatenating gene count and batch embedding as an dimension-expanded input. The new input is forwarded through torch.nn.Linear. This is not what we want to implement, right?

sjfleming:
Right, we’d like to do it differently so that the batch bias vector can be sampled from a distribution (and not just computed directly using a weight matrix).  There could be multiple ways to implement this, but I don’t think the current scvi implementation can do what we need

sjfleming:
`LinearWithBatch` now accepts an __init__ argument called `sample`, which is supposed to make this easier.  #156 has been merged to the `modified_scvi` branch, so please base your work on that branch @ImXman

Issue #144, opened 2024-03-20
scvi - use precomputed batch bias vectors


sjfleming:
We want the ability to use initialize the model with pre-computed batch bias vectors and use those (without updating them) throughout training.

We should set things up so that those batch bias vectors can be pulled from a checkpoint of a previously-run scVI model.  We should have checks in place to make sure that the checkpointed run is "compatible" with this run (i.e. same batches are present).

sjfleming:
See this example of how to use a checkpoint in a config file:
https://github.com/cellarium-ai/cellarium-ml/blob/module-tutorial/configs/lr_config.yaml#L47-L52

LysSanzMoreta:
@sjfleming  Just to make sure, this would be the affected part: https://github.com/cellarium-ai/cellarium-ml/blob/modified_scvi/cellarium/ml/models/common/nn.py#L29C1-L47C34

Right? Instead of computing them , use the ones given by the checkpoint

sjfleming:
Yeah, I think `LinearWithBatch` might be a good place to intervene to make this happen.

There are many ways, but maybe we can come up with a simple one...
what about something like:
- make a new input argument in `LinearWithBatch.__init__()`, like maybe `fix_batch_effect: bool` or something
- then use `self.fix_batch_effect` in, maybe, `self.compute_bias()`?  It could be something like `return self.cached_biases` if `self.fix_batch_effect` is false, and if it's true then `return self.cached_biases.detach()` ?
- there might be better ways...

I think we can let the checkpoint loader be responsible for making sure that `self.bias_mean_layer` for `LinearWithBatch` gets loaded correctly, so it should already have the pre-trained batch-biases.  I think the idea here is just to be able to turn off further training of the batch-biases.

PR #156, opened 2024-03-28
scvi: make encoder/decoder more flexible

### LINKED ISSUE #142: scvi - make encoder and decoder architecture flexible
scvi-tools has the encoder architecture fixed as the mirror of the decoder (unless you're using LinearlyDecodededSCVI).

We want to be able to specify the architecture (size, n_layers) of the two networks independently.

### LINKED ISSUE #143: scvi - flexibly inject batch covariates into VAE
We want the ability to inject the batch one-hot covariates at any layer of either the encoder or decoder.  Or multiple layers.

image

(Forget about the KL divergence in the picture above, just note that the batch covariates are being injected at different places.)

sjfleming:
These changes are aimed at making the VAE architecture for scVI totally flexible.  You can choose how many layers, their sizes, and whether or not they include a batch-specific bias term.  You can make these choices separately for encoder and decoder.  You can also include a batch-specific bias term on the output layers of the encoder and decoder (something scvi-tools does not allow).

Closes #142 
Closes #143 

Hopefully enables #144 and #145


sjfleming:
Currently this is working (except for the case of zero hidden layers in either encoder or decoder) and you can specify encoder and decoder architectures from the config file as 
```
      encoder_hidden: 
      - class_path: cellarium.ml.models.common.fclayer.LinearInputBias
        init_args:
          out_features: 128
      - class_path: torch.nn.Linear
        init_args:
          out_features: 64
      decoder_hidden:
      - class_path: cellarium.ml.models.common.fclayer.LinearInputBias
        init_args:
          out_features: 128
```

But I still have a question...

I am a bit confused about how scVI is supposed to work in the case of a linear decoder.  It seems like, if the decoder is linear, then we end up adding a batch-dependent bias vector $x_g^\text{(batch)}$ to the output data.
https://github.com/scverse/scvi-tools/blob/6dedd600ff25bd388726746d8696fb4af40eb125/scvi/module/_vae.py#L823
https://github.com/scverse/scvi-tools/blob/6dedd600ff25bd388726746d8696fb4af40eb125/scvi/nn/_base_components.py#L457
But if the decoder is not linear (i.e. there is at least one hidden layer), we do not add this $x_g^\text{(batch)}$ bias to the output.
https://github.com/scverse/scvi-tools/blob/6dedd600ff25bd388726746d8696fb4af40eb125/scvi/nn/_base_components.py#L356
Is that true, or am I missing something?  

(The scVI model could always add $x_g^\text{(batch)}$ to the output, and that would kind of be like regressing out the linear fixed effects of batch (and not having to model it using the VAE)... but I don't think this is what scVI is currently doing.)

sjfleming:
With 16-dimensional latent and a single 128-unit hidden layer in both encoder and decoder, I see
image

sjfleming:
The pink one includes a `Parameter` init in `.reset_parameters()`, but the orange-ish one does not.
image

LysSanzMoreta:
@sjfleming with respect to the batch mixing issue (assuming that you refer to the fact that with the Parameter initialization there is potentially an issue due to more separation according to the different experiments types and less otherwise):
i)  How many times does this phenomena occur? is it consistent?
ii) Could you try this weight initalization function in case there is some torch.module that has been forgotten to be initialized? Torch weight parameters are initalized to `U(− 0.07 , 0.07)` and `zero bias`. I think with this function you make sure you did not forget anything

```python
   def reset_parameters(self) -> None:
        self.apply(self.init_parameters)

    def init_parameters(self, m) -> None:
        if isinstance(m, torch.nn.Module) and hasattr(m, 'weight'):
            if isinstance(m, torch.nn.BatchNorm1d):
                torch.nn.init.normal_(m.weight, 1.0, 0.02)
            elif isinstance(m, torch.nn.Linear):
                torch.nn.init.xavier_normal_(m.weight)
            if m.bias is not None:
                torch.nn.init.zeros_(m.bias)
        elif isinstance(m,torch.nn.Parameter) and hasattr(m, 'weight'):
            torch.nn.init.xavier_normal_(m.weight)
            if m.bias is not None:
                torch.nn.init.zeros_(m.bias)
```

sjfleming:
Everything seems to run and work just as well with the code you suggested, but I think that may be because the `torch.nn.Parameter`s are still not being initialized.

I think that 
```python
isinstance(m, torch.nn.Parameter) and hasattr(m, 'weight')
```
will always return `False` because a `torch.nn.Parameter` does not have a `weight` attribute.

For example, if I put in a print statement
```python
    def init_parameters(self, m) -> None:
        if isinstance(m, torch.nn.Module) and hasattr(m, 'weight'):
            if isinstance(m, torch.nn.BatchNorm1d):
                torch.nn.init.normal_(m.weight, 1.0, 0.02)
            elif isinstance(m, torch.nn.Linear):
                torch.nn.init.xavier_normal_(m.weight)
            if m.bias is not None:
                torch.nn.init.zeros_(m.bias)
        elif isinstance(m,torch.nn.Parameter) and hasattr(m, 'weight'):
            print('init param')
            torch.nn.init.xavier_normal_(m.weight)
            if m.bias is not None:
                torch.nn.init.zeros_(m.bias)
```

I never see it print `init param`

sjfleming:
(Also...)

If I run this
```python
    def init_parameters(self, m) -> None:
        if isinstance(m, torch.nn.Module) and hasattr(m, 'weight'):
            if isinstance(m, torch.nn.BatchNorm1d):
                torch.nn.init.normal_(m.weight, 1.0, 0.02)
            elif isinstance(m, torch.nn.Linear):
                torch.nn.init.xavier_normal_(m.weight)
            if m.bias is not None:
                torch.nn.init.zeros_(m.bias)
        elif isinstance(m,torch.nn.Parameter):
            print('found a param')
            if hasattr(m, 'weight'):
                print('init param')
                torch.nn.init.xavier_normal_(m.weight)
                if m.bias is not None:
                    torch.nn.init.zeros_(m.bias)
```

I never see `found a param`... which is why I switched to this
```python
    def reset_parameters(self) -> None:
        for m in self.modules():
            m.apply(weights_init)
        for p in self.parameters():
            weights_init(p)
```
but I feel like I must be doing something wrong... seems like `self.apply(self.init_parameters)` should include the `torch.nn.Parameter`s in `self`...

sjfleming:
Oh, but what I had about 
```python
for p in self.parameters():
    weights_init(p)
```
is definitely not a good thing!  It doesn't just get the top level `Parameters`, it digs down to each of parameters of each sub-module... effectively initializing everything again!  Aahh

LysSanzMoreta:
This is my current proposal:

```python
import torch

class TestModule(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.p = torch.nn.Parameter(torch.randn(5))
        self.m = torch.nn.Linear(5, 10)
        #self.param_dict = dict(self.named_parameters())

    def reset_parameters(self) -> None:
        self.apply(self.init_parameters)
        for param in self.named_parameters():
            self.init_parameters(param)

    def init_parameters(self, m) -> None:

        print(m.__class__.__name__)
        print(m)
        if isinstance(m, torch.nn.Module) and hasattr(m, 'weight'):
            print("Found module")
            if isinstance(m, torch.nn.BatchNorm1d):
                torch.nn.init.normal_(m.weight, 1.0, 0.02)
            elif isinstance(m, torch.nn.Linear):
                torch.nn.init.xavier_normal_(m.weight)
            #missing an else case here
            if m.bias is not None:
                torch.nn.init.zeros_(m.bias)
        elif isinstance(m,tuple): #parameters come in tuple form
            if len(m) == 2:
                #param = self.param_dict[m[0]]
                param = m[1]
                if hasattr(param,'weight'):
                    torch.nn.init.xavier_normal_(param.weight)
                if hasattr(param,'bias'):
                    torch.nn.init.zeros_(param.bias)
            else:
                raise ValueError("If this case exists it needs to be addressed")
        else:
            print("Found nothing")


TestModule().reset_parameters()
```

It goes over the parameters only once

sjfleming:
Let's keep thinking about `reset_parameters()`... made an issue here for future work: #157

LysSanzMoreta (commented):
good

PR #158, opened 2024-04-04
reset_parameters update

ImXman:
All parameters should be initialized the same time when calling reset_parameters()

px_r is the only torch.nn.Parameters rather than a neural network. We could initialize px_r in weight_init. However, there are 3 options of initialization of px_r depending on the setup of dispersion. Moving initialization of px_r into weight_init makes the choice of initialization of px_r inflexible.

### DIFF
--- a/cellarium/ml/models/scvi.py
+++ b/cellarium/ml/models/scvi.py
@@ -356,19 +356,7 @@ class SingleCellVariationalInference(CellariumModel, PredictMixin):
             self.register_buffer("library_log_means", torch.from_numpy(library_log_means).float())
             self.register_buffer("library_log_vars", torch.from_numpy(library_log_vars).float())
 
-        if self.dispersion == "gene":
-            self.px_r = torch.nn.Parameter(torch.randn(self.n_input))
-        elif self.dispersion == "gene-label":
-            raise NotImplementedError
-            # self.px_r = torch.nn.Parameter(torch.randn(self.n_input, n_labels))
-        elif self.dispersion == "gene-cell":
-            self.px_r = torch.nn.Parameter(torch.zeros(1))  # dummy
-        else:
-            raise ValueError(
-                "dispersion must be one of ['gene', "
-                " 'gene-label', 'gene-cell'], but input was "
-                "{}".format(self.dispersion)
-            )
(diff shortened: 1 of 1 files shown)


cellarium/ml/models/scvi.py
```diff
@@ -356,19 +356,7 @@ def __init__(
...
-            # self.px_r = torch.nn.Parameter(torch.randn(self.n_input, n_labels))
-        elif self.dispersion == "gene-cell":
-            self.px_r = torch.nn.Parameter(torch.zeros(1))  # dummy
-        else:
-            raise ValueError(
-                "dispersion must be one of ['gene', "
-                " 'gene-label', 'gene-cell'], but input was "
-                "{}".format(self.dispersion)
-            )
+        self.px_r = torch.nn.Parameter(torch.torch.empty(self.n_input))
```
sjfleming, line 359: `torch.torch.empty(self.n_input)` will not work

cellarium/ml/models/scvi.py
```diff
@@ -445,7 +433,19 @@ def __init__(
...
+        elif self.dispersion == "gene-label":
+            raise NotImplementedError
+        elif self.dispersion == "gene-cell":
+            torch.nn.init.zeros_(self.px_r)
+        else:
+            raise ValueError(
+                "dispersion must be one of ['gene', "
+                " 'gene-label', 'gene-cell'], but input was "
+                "{}".format(self.dispersion)
+            )
```
sjfleming, line 448: I think we want to keep this stuff in `__init__()`.

`self.px_r` is only used if `self.dispersion == "gene"` or if `self.dispersion == "gene_label"`, but `self.px_r` is a different size in the two cases.  We don't want to be creating the parameter in `reset_parameters()`, we need to create parameters in `__init__()`.  So we have to have this if-statement in `__init__()` anyway (if we ever want "gene-label" to work).  I think it's easier to just have this here:
```python
torch.nn.init.normal_(self.px_r, mean=0.0, std=1.0)
```

sjfleming (commented):
I'm not sure about this one... I kind of like it the way it is now.  Let me know what you think in comments.

PR #166, opened 2024-04-12
Add example configs

### LINKED ISSUE #114: Add example config files
General comment: please make an `examples` directory (or somthing to that effect) and populate it with several useful config files, each with a separate README. For example:
- OnePassMeanVarStd -> Incremental PCA
- TDigest w/ lib size normalization + log transform -> LogisticRegression w/ lib size normalization + log transform + divide by median

_Originally posted by @mbabadi in https://github.com/cellarium-ai/cellarium-ml/pull/108#pullrequestreview-1830395954_

ordabayevy:
Resolves #114

### DIFF
--- a/examples/cli_workflow/README.md
+++ b/examples/cli_workflow/README.md
@@ -1,0 +1,288 @@
+# Example workflow
+
+This example workflow demonstrates how to:
+
+1. Compute the mean and standard deviation of normalized and log1p-transformed data.
+2. Perform PCA on data that has been z-scored based on the mean and standard deviation from step 1.
+3. Train a logistic regression classifier on the principal component embeddings of the data.
+
+To execute the workflow, run the scripts in sequence:
+
+```bash
+cellarium-ml onepass_mean_var_std fit --config onepass_train_config.yaml
+cellarium-ml incremental_pca fit --config ipca_train_config.yaml
+cellarium-ml logistic_regression fit --config lr_train_config.yaml
+```
+## `OnePassMeanVarStd`
+
+Generate a default config file:
+
+```
+cellarium-ml onepass_mean_var_std fit --print_config > onepass_train_config.yaml
+```
+
+Below we highlight the changes made to the default configuration file.
+
+### trainer
+
+Change the number of devices and set the path for logs and weights:
+
+```diff
+<   devices: auto
+<   default_root_dir: null
+---
+>   devices: 2
+>   default_root_dir: /tmp/test_examples/onepass
+```
+
+### model
+
+Add `NormalizeTotal` and `Log1p` transforms:
+
+```diff
+<   transforms: null
+---
+>   transforms:
+>     - class_path: cellarium.ml.transforms.NormalizeTotal
+>       init_args:
+>         target_count: 10_000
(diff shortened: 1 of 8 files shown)


examples/cli_workflow/README.md
```diff
@@ -0,0 +1,288 @@
...
+
+```diff
+<       algorithm: naive
+---
+>       algorithm: shifted_data
+```
+
+### data
+
+Configure the `DistributedAnnDataCollection`. Here we validate `obs` columns that are used by the transforms and the model (`total_mrna_umis`):
```
mbabadi, line 62: Say a word about what it means to "validate` an obs column.

examples/cli_workflow/README.md
```diff
@@ -0,0 +1,288 @@
...
+>           file_path: /tmp/test_examples/onepass/lightning_logs/version_0/checkpoints/epoch=0-step=2.ckpt
+>           attr: model.std_g
+>           convert_fn: null
+>         var_names_g:
+>           !CheckpointLoader
+>           file_path: /tmp/test_examples/onepass/lightning_logs/version_0/checkpoints/epoch=0-step=2.ckpt
+>           attr: model.var_names_g
+>           convert_fn: numpy.ndarray.tolist
+```
+
```
mbabadi, line 149: Maybe add a note below that cellarium-ml does not perform any validation on the content of data loaded from the checkpoints, or whether it is consistent with the rest of the configuration. For example, the mean and std were calculate in a prior stop from data that was subject to normalize total and log1p transform. If the user inadvertently forgets to perform the same transforms here before z-scoring, the workflow will run w/o any error though will produce wrong results.

examples/cli_workflow/README.md
```diff
@@ -0,0 +1,288 @@
...
+Set the number of components and mean correction option to true for `IncrementalPCA`:
+
+```diff
+<       n_components: null
+<       perform_mean_correction: false
+---
+>       n_components: 50
+>       perform_mean_correction: true
+```
+
```
mbabadi, line 159: Again, add a sticky note that since we have z-scored the data, mean correction is not strictly necessary but its presence may help mitigate roundoff errors. It is a good exercise for the user to ascertain that the learned mean by IPCA is close to 0/

examples/cli_workflow/README.md
```diff
@@ -0,0 +1,288 @@
...
+```
+cellarium-ml logistic_regression fit --print_config > lr_train_config.yaml
+```
+
+Below we highlight the changes made to the default configuration file.
+
+### train
+
+Change the number of devices, change strategy to `ddp_find_unused_parameters_true` (because trained PCA model contains parameters that are fixed during training), set the number of epochs, and set the path for logs and weights:
```
mbabadi, line 209: perhaps give a reference to PL doc related to this?

examples/cli_workflow/README.md
ordabayevy: I set`perform_mean_correction: true` by mistake. It should be `false` in this case because of the z-score!

mbabadi (approved):

PR #169, opened 2024-04-19
scvi - enable batch embeddings, sampling, and freezing of batch bias computations

### LINKED ISSUE #167: scvi - refactor to enable universal model description
Kind of a nebulous issue title.  What we want to do here is enable the following kinds of model experimentation:

- normal scvi-tools
- have "batch" be a random variable of some particular dimension that is sampled from its normal posterior
   - use this in computing "batch-biases" to be injected in layers
   - optionally promote "batch" to a latent variable and add a KL divergence from a standard normal
   - choose the dimension of this z_batch
- optionally add a batch-dependent linear bias to the final output reconstruction (after the nonlinearity): "beta"

Implementation:

- there is always a "z_batch" specified by a posterior mean and std
- sometimes this "z_batch" can be sampled from its posterior, sometimes it can be fixed at its mean (which would be normal scvi-tools)
- a KL divergence can be optionally applied if "z_batch" is sampled from its posterior
- "z_batch" gets projected to "z_batch_bias" vectors (for layers where we are conditioning) by means of layer-specific decoder NNs (flexible architecture but same for all layers)

A "normal" run of scvi-tools would correspond to a "z_batch" that does not get sampled (just use the posterior mean) coupled with nonlinear decoders of decent complexity, and no KL divergence on "z_batch".

Related to #145

sjfleming:
Closes #167 

This is a rather large refactor to change the way that batch-bias values are computed.  It is closer to the way scvi-tools does things, but instead of concatenating a one-hot batch label directly to the input data (for a given layer), this implementation uses a separate neural network (need not be linear, but can be) to compute batch-biases.  This allows things to be much more flexible.  For instance, the batch input can be one-hot, or some other representation of batch (sampled from a batch embedding space for example), and it allows us to easily freeze the network that computes the batch-bias values (if we don't want to train them, but rather use ones from a previous checkpoint).


sjfleming:
This does not yet address #144, but hopefully it will make it easier

sjfleming:
Currently, there is a new notion of a "batch embedding" which can be one-hot, or can be learned.  If it is learned, it can be a point estimate or it can be sampled from a posterior distribution.  If sampled, you can optionally add a KL divergence to the loss function.

ImXman (approved):
Tested

PR #171, opened 2024-04-22
NegativeBinomial parametrized by `mu` and `theta`

ordabayevy:
This is refactored from the `cellarium-gpt` branch. Special care taken of the case when `theta==0`.

### DIFF
--- a/cellarium/ml/distributions/negative_binomial.py
+++ b/cellarium/ml/distributions/negative_binomial.py
@@ -7,6 +7,8 @@ import torch
 from pyro.distributions import TorchDistribution, constraints
 from torch.distributions.utils import broadcast_all, lazy_property
 
+THETA_THRESHOLD_STIRLING_SWITCH = 200
+
 
 class NegativeBinomial(TorchDistribution):
     """Negative binomial distribution.
@@ -66,7 +68,7 @@ class NegativeBinomial(TorchDistribution):
         #       - torch.lgamma(value + 1)
         #   )
         delta = torch.where(
-            (value / self.theta < 1e-2) & (self.theta > 1e2),
+            self.theta > THETA_THRESHOLD_STIRLING_SWITCH,
             (value + self.theta - 0.5) * torch.log1p(value / self.theta) - value,
             (value + self.theta).lgamma() - self.theta.lgamma() - torch.xlogy(value, self.theta),
         )
--- a/cellarium/ml/distributions/negative_binomial.py
+++ b/cellarium/ml/distributions/negative_binomial.py
@@ -7,8 +7,6 @@ import torch
 from pyro.distributions import TorchDistribution, constraints
 from torch.distributions.utils import broadcast_all, lazy_property
 
-THETA_THRESHOLD_STIRLING_SWITCH = 200
-
 
 class NegativeBinomial(TorchDistribution):
     """Negative binomial distribution.
@@ -19,6 +17,8 @@ class NegativeBinomial(TorchDistribution):
         theta:
             Inverse dispersion.
     """
+
+    THETA_THRESHOLD_STIRLING_SWITCH = 200
 
     arg_constraints = {"mu": constraints.greater_than_eq(0), "theta": constraints.greater_than_eq(0)}
     support = constraints.nonnegative_integer
@@ -68,7 +68,7 @@ class NegativeBinomial(TorchDistribution):
         #       - torch.lgamma(value + 1)
         #   )
         delta = torch.where(
-            self.theta > THETA_THRESHOLD_STIRLING_SWITCH,
+            self.theta > self.THETA_THRESHOLD_STIRLING_SWITCH,
             (value + self.theta - 0.5) * torch.log1p(value / self.theta) - value,
             (value + self.theta).lgamma() - self.theta.lgamma() - torch.xlogy(value, self.theta),
         )
--- a/cellarium/ml/distributions/negative_binomial.py
+++ b/cellarium/ml/distributions/negative_binomial.py
@@ -1,0 +1,79 @@
+# Copyright Contributors to the Cellarium project.
+# SPDX-License-Identifier: BSD-3-Clause
+
+from numbers import Number
+
+import torch
+from pyro.distributions import TorchDistribution, constraints
+from torch.distributions.utils import broadcast_all, lazy_property
+
+
+class NegativeBinomial(TorchDistribution):
+    """Negative binomial distribution.
+
+    Args:
+        mu:
+            Mean of the distribution.
+        theta:
+            Inverse dispersion.
+    """
+
+    arg_constraints = {"mu": constraints.greater_than_eq(0), "theta": constraints.greater_than_eq(0)}
+    support = constraints.nonnegative_integer
+
+    def __init__(self, mu: torch.Tensor, theta: torch.Tensor, validate_args: bool | None = None) -> None:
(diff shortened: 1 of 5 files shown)


sjfleming:
Excellent, the scvi model would love to use this as well

cellarium/ml/distributions/negative_binomial.py
```diff
@@ -0,0 +1,79 @@
...
+        #       - torch.lgamma(self.theta)
+        #       - torch.lgamma(value + 1)
+        #   )
+        delta = torch.where(
+            (value / self.theta < 1e-2) & (self.theta > 1e2),
+            (value + self.theta - 0.5) * torch.log1p(value / self.theta) - value,
+            (value + self.theta).lgamma() - self.theta.lgamma() - torch.xlogy(value, self.theta),
+        )
+        # The case self.theta == 0 and value == 0 has probability 1.
+        # The case self.theta == 0 and value != 0 has probability 0.
```
sjfleming, line 74: I guess in this parameterization, based on mean and variance, `theta` is the "inverse overdispersion".  I am thinking about this edge case.

For the $\Phi$ parameterization where variance is $\mu + \Phi \mu^2$, the case $\Phi = 0$ reverts to a Poisson.  The case $\Phi \rightarrow \inf$, which I guess is this limit $\theta \rightarrow 0$... what does happen... I guess maybe you're right! ... does it become a delta function?  I'm not very familiar with that limit.  In some sense it kind of seems like we don't want our models to approach that limit.  Infinite overdispersion is never something we want a model to be able to land on, as it doesn't seem very sensible.  And it completely ignores the mean $\mu$ and kind of becomes a one-parameter distribution rather than a two-parameter distribution.

In the $\theta \rightarrow 0$ limit, it seems like $\mu$ becomes totally unconstrained.  I guess it is the responsibility of the modeler to make sure that $\theta$ is learned in such a way that it never actually goes to 0.

tests/distributions/test_negative_binomial.py
```diff
@@ -0,0 +1,54 @@
...
+import torch
+
+from cellarium.ml.distributions import NegativeBinomial
+
+
+@pytest.mark.parametrize("logits_shape", [(), (2,), (3, 2)])
+@pytest.mark.parametrize("total_counts_shape", [(), (2,), (3, 2)])
+def test_negative_binomial(logits_shape: torch.Size, total_counts_shape: torch.Size) -> None:
+    logits = torch.randn(logits_shape)
+    total_counts = torch.rand(total_counts_shape) * 10
```
sjfleming, line 16: Maybe it would also make sense to hit some edge cases, like the reversion to Poisson when $\theta \rightarrow \inf$, and things like small $\mu$ and extremely large $\theta$.

cellarium/ml/distributions/negative_binomial.py
```diff
@@ -0,0 +1,79 @@
...
+        #   log_theta_mu_eps = torch.log(self.theta + self.mu + self.eps)
+        #   return (
+        #       self.theta * (torch.log(self.theta + self.eps) - log_theta_mu_eps)
+        #       + value * (torch.log(self.mu + self.eps) - log_theta_mu_eps)
+        #       + torch.lgamma(value + self.theta)
+        #       - torch.lgamma(self.theta)
+        #       - torch.lgamma(value + 1)
+        #   )
+        delta = torch.where(
+            (value / self.theta < 1e-2) & (self.theta > 1e2),
```
sjfleming, line 69: I have to admit I do not understand how this works.  Can you explain the logic here @ordabayevy ?

mbabadi: @ordabayevy Actually we don't need the first condition-- my bad. Let's do this:

```
theta = torch.where(
    self.theta > THETA_THRESHOLD_STIRLING_SWITCH,
    (value + self.theta - 0.5) * torch.log1p(value / self.theta) - value,
    (value + self.theta).lgamma() - self.theta.lgamma() - torch.xlogy(value, self.theta),
)
```

and let's make `THETA_THRESHOLD_STIRLING_SWITCH = 200.` a class constant.

@sjfleming The idea is to swtich to leading order Stirling approximation for gamma when `self.theta` is large enough. If `self.theta` is large enough, then `value + self.theta` is also at least as large since `value >= 0`. The expression `(value + self.theta - 0.5) * torch.log1p(value / self.theta) - value` is indeed the leading order Stirling approximation of the exact expression `(value + self.theta).lgamma() - self.theta.lgamma() - torch.xlogy(value, self.theta)`, see [here](https://en.wikipedia.org/wiki/Stirling%27s_approximation#Stirling's_formula_for_the_gamma_function).

tests/distributions/test_negative_binomial.py
mbabadi: Let's make another test and sweep `mu` in `[1e-4, ..., 1e3]` and `theta` in `[1e-2, ..., 1e3]` calculate the total probability (via logsumexp over a range of `value` in `[0, ..., 1e3 * mu]`) and assert that it is close to `1.0` with a tolerance of `1e-5` using `float32`.

mbabadi (changes requested):
Nicely done, a few small suggestions.

cellarium/ml/distributions/negative_binomial.py
sjfleming: Very interesting!

cellarium/ml/distributions/negative_binomial.py
ordabayevy: I've been following the implementation of Negative Binomial from PyTorch Distributions: https://pytorch.org/docs/stable/_modules/torch/distributions/negative_binomial.html#NegativeBinomial

cellarium/ml/distributions/negative_binomial.py
ordabayevy: @mbabadi sounds good! I have to admit that I didn't dive deep into how this approximation works but I trust you!

tests/distributions/test_negative_binomial.py
ordabayevy: Added test, however, it only passes if I set `atol=5e-4`

sjfleming: Like, is there a plot of y-axis `THETA_THRESHOLD_STIRLING_SWITCH`, x-axis `atol` that could be created?

ordabayevy: > Does the necessary `atol` value change if you increase `THETA_THRESHOLD_STIRLING_SWITCH` to like 500 or 1000 ?

Just tried it, it didn't. My understanding from that wiki article that this approximation breaks down only for very small theta.

ordabayevy:
@mbabadi added the test that you suggested. Had to increase the range of theta to get into cases where tests would fail without the Stirling approximation. Here are some comparisons of the `delta` values as a function of `theta` that I tried:

```py
value = 10
theta = torch.logspace(0, 6, 100)
delta1 = (value + theta - 0.5) * torch.log1p(value / theta) - value
delta2 = (value + theta).lgamma() - theta.lgamma() - torch.xlogy(value, theta)

plt.figure(figsize=(4, 4))
plt.plot(theta.log(), delta1, label="stirling approx")
plt.plot(theta.log(), delta2, label="exact")
plt.xlabel("log(theta)")
plt.ylabel("delta")
plt.legend()
```

image

mbabadi (approved):

PR #173, opened 2024-04-23
Improve error messages

### LINKED ISSUE #95: Improve error messages

ordabayevy:
Resolves #95 . It is a better practice to raise ValueError instead of using assertions. Use assertion only in tests and for type narrowing.


sjfleming (approved):
Looks good to me!

mbabadi (approved):
Excellent!

PR #174, opened 2024-04-25
Fix README

ordabayevy:
Uploading to TestPyPI fails with the following message:

```
Checking dist/cellarium_ml-0.0.4.post12-py3-none-any.whl: FAILED
ERROR    `long_description` has syntax errors in markup and would not be        
         rendered on PyPI.                                                      
         line 19: Error: Unexpected indentation. 
```

This PR fixes indentation issue in the README file.


sjfleming (approved):
Nice!  Had no idea that a testpypi upload could fail for that reason.

PR #195, opened 2024-05-31
Implement contrastive learning model and transforms

bricewang:
Added contrastive multilayer perceptron and NT cross-entropy loss

Added Duplicate, BinomialResample, Dropout, Gaussian noise transforms

Added contrastive_mlp to CLI


cellarium/ml/models/__init__.py
```diff
@@ -1,22 +1,26 @@
 # Copyright Contributors to the Cellarium project.
 # SPDX-License-Identifier: BSD-3-Clause
 
+from cellarium.ml.models.contrastive_mlp import ContrastiveMLP
 from cellarium.ml.models.geneformer import Geneformer
 from cellarium.ml.models.incremental_pca import IncrementalPCA
 from cellarium.ml.models.logistic_regression import LogisticRegression
 from cellarium.ml.models.model import CellariumModel, PredictMixin, ValidateMixin
 from cellarium.ml.models.mu_linear import MuLinear, abcdParameter
+from cellarium.ml.models.nt_xent import NT_Xent
```
ordabayevy, line 10: Refactor: I suggest having a `cellarium.ml.losses` folder and add our losses there.

cellarium/ml/models/contrastive_mlp.py
```diff
@@ -0,0 +1,99 @@
...
+            Size of embedding (network output size).
+        world_size:
+            Number of devices used in training.
+        temperature:
+            Parameter governing Normalized Temperature-scaled cross-entropy (NT-Xent) loss.
+    """
+
+    def __init__(
+        self,
+        g_genes: int,
```
ordabayevy, line 34: Naming convention: at some point we switched from `g_genes` naming convention to `n_obs`. You can also switch it to `n_obs` to be consistent with other models.

cellarium/ml/models/contrastive_mlp.py
```diff
@@ -0,0 +1,99 @@
...
+        hidden_size: Sequence[int],
+        embed_dim: int,
+        world_size: int,
+        temperature: float = 1.0,
+    ):
+        super(ContrastiveMLP, self).__init__()
+
+        layer_list: List[nn.Module] = []
```
ordabayevy, line 42: Minor comment: you can create an empty `nn.Sequential()` and append to it directly:

```py
self.layers = nn.Sequential()
self.layers.append(module)
```

cellarium/ml/models/contrastive_mlp.py
```diff
@@ -0,0 +1,99 @@
...
+        z = F.normalize(self.layers(x_ng))
+
+        # split input into augmented halves
+        z1, z2 = torch.chunk(z, 2)
+
+        # SimCLR loss
+        loss = self.Xent_loss(z1, z2)
+        return {"loss": loss}
+
+    def predict(self, x_ng: torch.Tensor, **kwargs: Any):
```
ordabayevy, line 87: Minor comment: `**kwargs` can be removed, right?

cellarium/ml/transforms/binomial_resample.py
```diff
@@ -0,0 +1,50 @@
...
+    def forward(self, x_ng: torch.Tensor) -> dict[str, torch.Tensor]:
+        """
+        Args:
+            x_ng: Gene counts.
+
+        Returns:
+            Binomially resampled gene counts.
+        """
+        p_binom_ng = Uniform(self.p_binom_min, self.p_binom_max).sample(x_ng.shape).type_as(x_ng)
+        p_apply_n = Bernoulli(probs=self.p_apply).sample(x_ng.shape[:1]).type_as(x_ng).bool()
```
ordabayevy, line 45: Maybe rename `p_apply_n` to something else, maybe like `apply_mask_n`? Because it is not a probability anymore.

cellarium/ml/transforms/binomial_resample.py
```diff
@@ -0,0 +1,50 @@
...
+        y_{ng} = Binomial(n=x_{ng}, p=p_{ng})
+
+    Args:
+        p_binom_min:
+            Lower bound on binomial distribution parameter.
+        p_binom_max:
+            Upper bound on binomial distribution parameter.
+    """
+
+    def __init__(self, p_binom_min, p_binom_max, p_apply):
```
ordabayevy, line 29: Add type hints to `__init__`.

cellarium/ml/transforms/binomial_resample.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def forward(self, x_ng: torch.Tensor) -> dict[str, torch.Tensor]:
+        """
+        Args:
+            x_ng: Gene counts.
+
+        Returns:
+            Binomially resampled gene counts.
+        """
+        p_binom_ng = Uniform(self.p_binom_min, self.p_binom_max).sample(x_ng.shape).type_as(x_ng)
```
ordabayevy, line 44: Are you using `.type_as` here to change the device?

`Uniform` uses the device of the `self.p_binom_min`. I think if it is of `float` type then it might be converted to `torch.Tensor` and be on CPU. Then `Uniform.sample` will sample on CPU which might be slow. In this case it might be better to convert `self.p_binom_min` to tensor yourself and cast to GPU device so that sampling step will be faster.

cellarium/ml/transforms/randomize.py
```diff
@@ -0,0 +1,35 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+
+
+import torch
+from torch import nn
+
+
+class Randomize(nn.Module):
+    """
+    Randomizely applies transform with probability p;
```
ordabayevy, line 11: Randomizely -> Randomly?

ordabayevy (changes requested):
Looks great! Left my initial comments. Can you also add the test in `test_cli::test_cpu_multi_device` and a loading from a checkpoint test similar to one in https://github.com/cellarium-ai/cellarium-ml/blob/main/tests/test_geneformer.py

cellarium/ml/transforms/binomial_resample.py
bricewang: Also fixed a bug where `x_aug` and `x_ng` were reversed in `torch.where`

cellarium/ml/transforms/randomize.py
bricewang: Alternatively I removed the entire class, as implementation constraints dictated that I determine whether or not to apply transform in the transform itself (i.e. `p_apply`). In future, this logic could also be pulled out into some inheritable entity.

Issue #175, opened 2024-04-25
Add a new `ParallelStrategy` that doesn't require gradient synchronization


ordabayevy:
At the moment we use `DDPStrategy` to train models that don't require gradient synchronization (`IncrementalPCA`, `OnePassMeanVarStd`). However, ddp strategy requires that your model has parameters which is why we have `_dummy_param` in these models. This works but it get's a bit inconvenient when the trained model is used as a transform because it becomes and unused parameter and ddp complains about it.

ordabayevy:
Starting from PyTorch Lightning 2.3 returning None in a training step is not allowed https://github.com/Lightning-AI/pytorch-lightning/pull/19918

PR #206, opened 2024-07-03
`DDPNoParametersStrategy`

ordabayevy:
- Remove the default strategy for `OnePassMeanVarStd` and `IncrementalPCA`
- Disable a LightningCLI breaking change introduced in PyTorch Lightning 2.3
- Add `sync_dist=True` flag to train loss log
- Add a new `DDPNoParametersStrategy` #175 
- In `TDigest` use `torchdistributed.gather_object` instead of writing/reading tdigest objects from files


ordabayevy:
Closed in favor of #206

PR #207, opened 2024-07-03
Support PyTorch Lightning 2.3

ordabayevy:
Fixes:
- Returning `None` from training step is not allowed during distributed training https://github.com/Lightning-AI/pytorch-lightning/pull/19918
- Breaking change introduced by the support for dependency injection (we instead rely on initialization under meta device context) https://github.com/Lightning-AI/pytorch-lightning/pull/18105


mbabadi (approved):
Nice!

PR #208, opened 2024-07-03
Distributed fixes & improvements

ordabayevy:
- Add `sync_dist=True` flag to train loss log
- In `TDigest` use `torchdistributed.gather_object` instead of writing/reading tdigest objects from files
- Move `GatherLayer` and `get_rank_and_num_replicas` to `cellarium.ml.utilities.distributed`


mbabadi (approved):
Nice!

PR #214, opened 2024-07-12
Remove reduntant resolve_ckpt method

ordabayevy:
Following up to #208 by removing the redundant `_resolve_ckpt_dir` method


mbabadi (changes requested):
This seems to be a useful method.  How about moving it to a utility class instead?

ordabayevy:
Good point. I added it to `cellarium.ml.utilities.core`.

mbabadi (approved):

Issue #204, opened 2024-06-26
scvi: include the option to limit to a specific set of genes


sjfleming:
We want a way to on-the-fly subset to a specific list of genes.

We do not want the dataloader to load the entire set of genes because this would be a big waste of memory / cuda memory.

Should this be a transform?  It seems like a transform happens too late... I think we'd really like this to be part of the dataloader itself.  Any thoughts @ordabayevy ?

What if we just wrote a new `convert_fn` for `X` to be used here in the config file?
```yaml
data:
  dadc:
    class_path: cellarium.ml.data.DistributedAnnDataCollection
    init_args:
      ...
  batch_keys:
    x_ng:
      attr: X
      convert_fn: cellarium.ml.utilities.data.subset_genes_and_densify
```

ordabayevy:
There is a [Filter transform](https://cellarium-ai.github.io/cellarium-ml/transforms.html#cellarium.ml.transforms.Filter) but like you say it might be too late if you worry about memory/speed. Yes, then the other place to add filtering before data got moved to gpu would be `convert_fn`. For that we probably need to figure out how to provide the list of subset genes.

sjfleming:
@ordabayevy These two commits provide a (working) rough idea.  But I don't know if the changes to core `data.py` are clean enough.  It could be improved.

https://github.com/cellarium-ai/cellarium-ml/commit/e21e7d9ddb2abe1055199c43acc0848168b144c7
https://github.com/cellarium-ai/cellarium-ml/commit/a7ed6c78cec520f03ed0bc7a0f600109ea3ecda9

In particular this config file
https://github.com/cellarium-ai/cellarium-ml/commit/a7ed6c78cec520f03ed0bc7a0f600109ea3ecda9

is ugly because of the repetition that would (seemingly?) be needed for X and var_names_g.  Hopefully we can find a more elegant way to handle it.  The referenced csv file just contains a list of gene names delimited by \n with no header.

ordabayevy:
I think adding `convert_fn_kwargs` is a nice extension! But I agree that additional logic in `data.py` is not very elegant. It almost feels like there might be a need for additional transforms before the data is moved to device 🤔 . Here are multiple ideas:

1. How often will the list of subset genes change? If it doesn't change much will it be easier to create a new extract files with filtered genes?
[...]

PR #220, opened 2024-07-26
Transforms split into before and after batch transfer

sjfleming:
Addresses the issue discussed in #204 , so that `Filter` (or any transform) can be applied on CPU before data is transferred to GPU.

Transforms are now implemented in two categories:
1. those that happen before transfer to device
2. those that happen after transfer to device

The config file will need to change a bit: instead of 
```
model:
  transforms:
    - class_path: cellarium.ml.transforms.NormalizeTotal
      init_args:
        target_count: 10_000
    - cellarium.ml.transforms.Log1p
```
you will need to write
```
model:
  after_batch_transfer_transforms:
    - class_path: cellarium.ml.transforms.NormalizeTotal
      init_args:
        target_count: 10_000
    - cellarium.ml.transforms.Log1p
```
and optionally you could do something like
```
model:
  before_batch_transfer_transforms:
    - class_path: cellarium.ml.transforms.Filter
      init_args:
        filter_list: ['gene1', 'gene2', 'gene3']
  after_batch_transfer_transforms:
    - class_path: cellarium.ml.transforms.NormalizeTotal
      init_args:
        target_count: 10_000
    - cellarium.ml.transforms.Log1p
```

Training now makes use of pytorch lightning hooks `on_before_batch_transfer()` and `on_after_batch_transfer()` to apply all transforms before the `batch` is passed to the `training_step()`.  The `training_step()` now only calls `forward()` on the model, and is not responsible for applying the transforms.

There are methods implemented in `CellariumPipeline` which still allow an entire pipeline to be applied to a dataset, say from the context of an interactive jupyter notebook.

### DIFF
--- a/cellarium/ml/cli.py
+++ b/cellarium/ml/cli.py
@@ -202,13 +202,19 @@ def compute_y_categories(data: CellariumAnnDataDataModule) -> np.ndarray:
     return field(adata)
 
 
-def compute_var_names_g(transforms: list[torch.nn.Module], data: CellariumAnnDataDataModule) -> np.ndarray:
+def compute_var_names_g(
+    before_batch_transfer_transforms: list[torch.nn.Module],
+    after_batch_transfer_transforms: list[torch.nn.Module],
+    data: CellariumAnnDataDataModule,
+) -> np.ndarray:
     """
     Compute variable names from the data by applying the transforms.
 
     Args:
-        transforms:
-            A list of transforms.
+        before_batch_transfer_transforms:
+            A list of transforms to be applied before batch transfer.
+        after_batch_transfer_transforms:
+            A list of transforms to be applied after batch transfer.
         data:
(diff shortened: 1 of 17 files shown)


sjfleming:
I think `crick` is causing an error unrelated to this PR's changes.  It might have to do with this?
https://stackoverflow.com/questions/40845304/runtimewarning-numpy-dtype-size-changed-may-indicate-binary-incompatibility

sjfleming:
@ordabayevy I made changes to `pyproject.toml` requirements (pinning `crick` and specifying an older `numpy`) solely to allow tests to succeed through github actions.  I don't know if these changes are a good idea.

But currently, a fresh `make install` of cellarium-ml will not result in a working installation, due to the problem with the `crick` installation binaries.

sjfleming:
For now, I am closing this in favor of #223

PR #223, opened 2024-08-01
CPU transforms on dataloader workers

sjfleming:
Small modifications to allow some optional set of transforms to be performed on CPU by dataloader workers.  Main use case would be the `Filter` transform, but it's possible that other transforms like `NormalizeTotal` and `Log1p` are so fast that they can be performed on CPU by dataloader workers as well, freeing up the GPU to just crank away on the model.

This adds the input `cpu_transforms` to `CellariumModule`.  If a `CellariumModule` is being used in the context of pytorch lightning training (i.e. if it `hasattr(self, "trainer")`) and if it is using a datamodule that is a `CellariumAnnDataDataModule`, then it will exclude the `cpu_transforms` from its list of applied transforms when calling `training_step()`, `forward()`, and `validation_step()`.


sjfleming:
@ordabayevy This is ready to take a look at, I think.  There's something real ugly going on with a `try / except` at the end of `CellariumModule.cofigure_model()`, but other than that, I think I like it.  Not sure what's the right way to check for a `Trainer`.

cellarium/ml/core/module.py
```diff
@@ -114,17 +153,50 @@ def configure_model(self) -> None:
...
+        self.pipeline = CellariumPipeline(cpu_transforms + transforms + (model,))  # the full pipeline
+        self.module_pipeline = self.pipeline  # training loop pipeline: initially the full pipeline
 
         if not self.hparams["is_initialized"]:
             model.reset_parameters()
             self.hparams["is_initialized"] = True
 
+        # move the cpu_transforms to the dataloader's collate_fn if the dataloader is going to apply them
+        try:
+            if hasattr(self, "trainer"):  # for some reason this check raises a RuntimeError rather than returning False
```
ordabayevy, line 169: We can maybe use private property `self._trainer` and judging from https://github.com/Lightning-AI/pytorch-lightning/issues/17517#issuecomment-1532325528 I think it is unlikely they will change it.

cellarium/ml/core/module.py
```diff
@@ -114,17 +153,50 @@ def configure_model(self) -> None:
...
+                if hasattr(self.trainer, "datamodule"):
```
ordabayevy, line 170: Seems it is safe to use `trainer.datamodule` per our conversation.

cellarium/ml/models/logistic_regression.py
```diff
@@ -104,11 +104,11 @@ def forward(
     def model(self, x_ng: torch.Tensor, y_n: torch.Tensor) -> None:
         W_gc = pyro.sample(
             "W",
-            dist.Laplace(0, self.W_prior_scale).expand([self.n_vars, self.n_categories]).to_event(2),
+            dist.Laplace(0, self.W_prior_scale).expand([self.n_vars, self.n_categories]).to_event(2),  # type: ignore[attr-defined]
```
ordabayevy, line 107: Are these type ignores necessary? The mypy version is currently pinned to 1.10.0 and type checking passed in the previous PR.

ordabayevy (changes requested):
This is really awesome @sjfleming !

One more thing that we should think through is what happens when you load `CellariumModule` from the checkpoint. One scenario is when resume training from a checkpoint using `pl.Trainer` and another scenario is when it is loaded by the user using `load_from_checkpoint`. Will `self._lightning_training_using_datamodule` work as expected and does it need to be persistent?

cellarium/ml/core/datamodule.py
```diff
@@ -126,6 +126,8 @@ def setup(self, stage: str | None = None) -> None:
             :attr:`val_dataset` is not shuffled and uses the ``same_order`` iteration strategy.
 
         """
+        self.collate_fn = collate_fn
```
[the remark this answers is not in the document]
sjfleming: Will do

cellarium/ml/core/module.py
sjfleming: Will do

cellarium/ml/core/module.py
sjfleming: Here I think I may have to keep the `hasattr(self.trainer, "datamodule")` check for mypy to pass.  Without it, for example if I try
```python
        # move the cpu_transforms to the dataloader's collate_fn if the dataloader is going to apply them
        if self._trainer is not None:
            # if hasattr(self.trainer, "datamodule"):
            if isinstance(self.trainer.datamodule, CellariumAnnDataDataModule):
                self._lightning_training_using_datamodule = True
                self.trainer.datamodule.collate_fn = FunctionComposer(
                    first_applied=collate_fn,
                    second_applied=self.cpu_transforms,
                )
                self.remove_cpu_transforms_from_module_pipeline()
```

I get this from mypy:
```pytb
error: "Trainer" has no attribute "datamodule"  [attr-defined]
```

cellarium/ml/core/module.py
```diff
@@ -3,30 +3,51 @@
...
+    """
+    Compose two functions into a single callable, in a way that is picklable.
+    """
+
+    def __init__(self, first_applied: Callable, second_applied: Callable):
+        self.first_applied = first_applied
+        self.second_applied = second_applied
+
+    def __call__(self, batch):
+        return self.second_applied(self.first_applied(batch))
```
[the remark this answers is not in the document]
sjfleming: Will do

cellarium/ml/models/logistic_regression.py
sjfleming: Well, if it passed before, it should be safe for me to remove these

cellarium/ml/core/module.py
ordabayevy: I think you can use either hasattr:

```py
if hasattr(self.trainer, "datamodule") and isinstance(self.trainer.datamodule, CellariumAnnDataDataModule):
```

or type ignore + walrus operator (to avoid additional type ignore)

```py
if isinstance(datamodule := self.trainer.datamodule, CellariumAnnDataDataModule):   # type: ignore[attr-defined]
    self._lightning_training_using_datamodule = True
    datamodule.collate_fn = FunctionComposer(
          first_applied=collate_fn,
          second_applied=self.cpu_transforms,
    )
```

Either should work.

ordabayevy:
Let's consider the scenario when you load the module from the checkpoint in the Jupyter Notebook:

`load_from_checkpoint` does three things:
1. Initialize `CellariumModule` using `hparams` in the checkpoint. At this step `self._lightning_training_using_datamodule` will be set to `False`.
2. Call `configure_model`. At this step the module will not detect the trainer and will keep `self._lightning_training_using_datamodule=False` and will not dispatch `cpu_transforms` to the datamodule.
3. Call `load_state_dict`. Nothing interesting happens here.

So in this scenario `cpu_transforms` will be run by the `CellariumModule`.

ordabayevy:
If you use `.fit(..., ckpt_path=ckpt_path)` or `.predict(..., ckpt_path=ckpt_path)` then the steps are:

1. `CellariumModule` is initialized by the user or from the config file. At this step `self._lightning_training_using_datamodule=False` and `configure_model` hasn't been called yet.
2. The module gets attached to the trainer.
3. Call `configure_model`. At this step `cpu_transforms` is dispatched to datamodule and `self._lightning_training_using_datamodule=True`
4. Call `load_state_dict`. Nothing interesting.

This case is similar to calling `.fit(...)` without `ckpt_path` argument.

ordabayevy (approved):
LGTM!

I think it can be merged unless @mbabadi you have additional comments.

sjfleming:
I should probably add a test to ensure that what you say is happening.  I don't think I am explicitly checking yet for all of those cases.

sjfleming:
Hm, I think there is a problem I've caught actually... I will update the code

ordabayevy:
A solution can be to have `module_pipeline` and `datamodule_pipeline` as a property (with setter probably). Then it might not show up in `CellariumModule`s `children()`. It should only save `pipeline` in the checkpoint.

sjfleming:
Expanded `test_cpu_transforms` to cover checkpoint loading, both manually and via `Trainer.fit(..., ckpt_path)`.  Getting the additional tests to pass did require making `module_pipeline` a `@property` of `CellariumModule`, rather than an object attribute.

cellarium/ml/core/module.py
```diff
@@ -274,3 +368,8 @@ def on_train_batch_end(self, outputs: STEP_OUTPUT, batch: Any, batch_idx: int) -
         on_batch_end = getattr(self.model, "on_batch_end", None)
         if callable(on_batch_end):
             on_batch_end(self.trainer)
+
+    # def on_load_checkpoint(self, checkpoint: torch.Dict[str, Any]) -> None:
+    #     print('here!')
+    #     print(checkpoint)
+    #     return super().on_load_checkpoint(checkpoint)
```
ordabayevy, line 375: Forgot to remove this.

cellarium/ml/core/module.py
```diff
@@ -62,6 +71,8 @@ def __init__(
             warnings.filterwarnings("ignore", message="Attribute 'model' is an instance of `nn.Module`")
             self.save_hyperparameters(logger=False)
         self.pipeline: CellariumPipeline | None = None
+        self._lightning_training_using_datamodule: bool = False
+        self._cpu_transforms_in_module_pipeline: bool = not self._lightning_training_using_datamodule
```
ordabayevy, line 75: Can we leave only one boolean flag here? Having two flags is unnecessary and can potentially lead to a bug. Then `remove_cpu_transforms_from_module_pipeline` can be removed as well because  setting `self._lightning_training_using_datamodule = True` essentially does it. Or if you want to remove `self._lightning_training_using_datamodule` and keep `remove_cpu_transforms_from_module_pipeline` then make it a private method because it should only be called by the `configure_model` and not by the user directly.

sjfleming: We can, I just am not sure what to call the flag.  I wanted to call it `_lightning_training_using_datamodule`, but then if I let the flag be set by callable methods, it's no longer guaranteed to be what the name implies.

Maybe I'll remove those helper methods and just act directly on `_lightning_training_using_datamodule`.

ordabayevy: Either name is fine, `_cpu_transforms_in_module_pipeline` sounds slightly better to me :)

> Maybe I'll remove those helper methods

Yeah, I agree. The flag is changed only in one place so having an entire method to do it might be too much.

sjfleming: Okay I'll change it to `_cpu_transforms_in_module_pipeline` which is more helpful in terms of "what the flag does".  `_lightning_training_using_datamodule` is more helpful in terms of what sets this flag, but eh, maybe less readable and less useful.

ordabayevy:
@sjfleming is it okay if I push the changes related to `compute_var_names_g`?

sjfleming:
Thanks for catching this @ordabayevy !
I'll add the changes now

sjfleming:
@ordabayevy I am seeing some unexpected behavior that's leading me to implement it using this inside `compute_var_names_g()`:
```python
    if (not cpu_transforms) and (not transforms):
        return batch["var_names_g"]
    pipeline = CellariumPipeline((cpu_transforms if cpu_transforms else []) + (transforms if transforms else []))
```

rather than this
```python
    pipeline = CellariumPipeline(cpu_transforms) + CellariumPipeline(transforms)
```

The issue is that the second implementation complains that "forward" is not implemented.  Here's the minimal demo:
```python
>>> from cellarium.ml import CellariumPipeline
>>> m = CellariumPipeline(None) + CellariumPipeline(None)
>>> m.forward()
```
```pytb
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/sfleming/miniconda3/envs/cellarium/lib/python3.10/site-packages/torch/nn/modules/module.py", line 374, in _forward_unimplemented
    raise NotImplementedError(f"Module [{type(self).__name__}] is missing the required \"forward\" function")
NotImplementedError: Module [ModuleList] is missing the required "forward" function
```
but, paradoxically (at least to me):
```python
>>> from cellarium.ml import CellariumPipeline
>>> m = CellariumPipeline(None)
>>> m.forward(None)
```
runs fine, and finds the "forward" method.

I think this points at a bug, where we really need to override `__add__` in `CellariumPipeline` or something?  Right now, it seems that the behavior of adding `CellariumPipeline`s is not properly handled.  Maybe a separate issue?

ordabayevy:
Sounds good.

> I think this points at a bug, where we really need to override __add__ in CellariumPipeline or something? Right now, it seems that the behavior of adding CellariumPipelines is not properly handled. Maybe a separate issue?

Agree

cellarium/ml/core/module.py
```diff
@@ -10,23 +10,31 @@
...
+        * :attr:`module_pipeline`: A :class:`cellarium.ml.core.CellariumPipeline` to apply all transforms,
+          minus the CPU transforms if they are handled by a :class:`CellariumAnnDataDataModule`, and the model.
         * :attr:`model`: A :class:`cellarium.ml.models.CellariumModel` to train with
           :meth:`training_step` method and epoch end hooks.
         * :attr:`optim_fn` and :attr:`optim_kwargs`: A Pytorch optimizer class and its keyword arguments.
         * :attr:`scheduler_fn` and :attr:`scheduler_kwargs`: A Pytorch lr scheduler class and its
           keyword arguments.
 
     Args:
+        cpu_transforms:
```
mbabadi, line 34: Please write some notes about the logic behind CPU transforms under a "Notes" section at the end of the docstring.

cellarium/ml/core/module.py
```diff
@@ -114,17 +139,45 @@ def configure_model(self) -> None:
                 )
             )
         else:
-            transforms = None
+            transforms = tuple()
 
-        self.pipeline = CellariumPipeline(transforms)
         if model is None:
```
mbabadi, line 144: Why check for `None` and not `~isinstance(model, CellariumModel)`?

ordabayevy: Yeah, it makes more sense to change it to `not isinstance(model, CellariumModel)`.

PR #228, opened 2024-08-07
Handle addition of `CellariumPipeline`s

### LINKED ISSUE #227: Handle methods when CellariumPipelines are added
Minimal demo:
```python
>>> from cellarium.ml import CellariumPipeline
>>> m = CellariumPipeline(None) + CellariumPipeline(None)
>>> m.forward()
```
```pytb
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/sfleming/miniconda3/envs/cellarium/lib/python3.10/site-packages/torch/nn/modules/module.py", line 374, in _forward_unimplemented
    raise NotImplementedError(f"Module [{type(self).__name__}] is missing the required \"forward\" function")
NotImplementedError: Module [ModuleList] is missing the required "forward" function
```
but this is fine as expected:
```python
>>> from cellarium.ml import CellariumPipeline
>>> m = CellariumPipeline(None)
>>> m.forward(None)
```
(runs fine)

I think this points at a bug, where we really need to override `__add__` in `CellariumPipeline` or something?  Right now, it seems that the behavior of adding `CellariumPipeline`s is not properly handled.

_Originally posted by @sjfleming in https://github.com/cellarium-ai/cellarium-ml/issues/223#issuecomment-2273843347_

sjfleming:
Closes #227


sjfleming:
This will need to wait for changes from #223 to pass mypy tests.

sjfleming:
I may also streamline `compute_var_names_g()` in cli.py once #223 is merged, because after this fix, we should be able to write
```python
CellariumPipeline(cpu_transforms) + CellariumPipeline(transforms)
```

cellarium/ml/core/pipeline.py
```diff
@@ -1,8 +1,12 @@
 # Copyright Contributors to the Cellarium project.
 # SPDX-License-Identifier: BSD-3-Clause
 
+from itertools import chain
+from typing import Iterable
```
ordabayevy, line 5: This should be `from collections.abc import Iterable` (for Python >= 3.9)

ordabayevy (approved):
🎉

PR #237, opened 2024-09-12
Fix Index import for the new version of anndata

ordabayevy:
In the new version of anndata `Index` and  `Index1D` need to be imported directly from their original module `anndata.compat`. We were importing them from `anndata._core.index` which imports from `anndata.compat`. But in the new version of anndata `TYPE_CHECKING` clause was added before the import statement.


sjfleming (approved):
Looks fine to me!  Is this change compatible with the older version of anndata, or will we need to update the anndata version requirement?

ordabayevy:
It should be compatible with the older version of anndata.

PR #251, opened 2024-10-09
Add `cellarium.ml.layers` module

ordabayevy:
This factors out varying layers from the CellariumGPT PR.

### DIFF
--- a/cellarium/ml/layers/__init__.py
+++ b/cellarium/ml/layers/__init__.py
@@ -1,0 +1,20 @@
+# Copyright Contributors to the Cellarium project.
+# SPDX-License-Identifier: BSD-3-Clause
+
+from cellarium.ml.layers.attention import MultiHeadAttention
+from cellarium.ml.layers.embedding import GeneEmbedding, MetaDataEmbedding
+from cellarium.ml.layers.ffn import PositionWiseFFN
+from cellarium.ml.layers.head import MultiHeadReadout
+from cellarium.ml.layers.normadd import NormAdd
+from cellarium.ml.layers.transformer import Transformer, TransformerBlock
+
+__all__ = [
+    "GeneEmbedding",
+    "MetaDataEmbedding",
+    "MultiHeadAttention",
+    "MultiHeadReadout",
+    "NormAdd",
+    "PositionWiseFFN",
+    "Transformer",
+    "TransformerBlock",
+]
--- a/cellarium/ml/layers/__init__.py
+++ b/cellarium/ml/layers/__init__.py
@@ -2,15 +2,15 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 from cellarium.ml.layers.attention import MultiHeadAttention
-from cellarium.ml.layers.embedding import GeneEmbedding, MetaDataEmbedding
+from cellarium.ml.layers.embedding import GeneExpressionEmbedding, MetadataEmbedding
 from cellarium.ml.layers.ffn import PositionWiseFFN
 from cellarium.ml.layers.head import MultiHeadReadout
 from cellarium.ml.layers.normadd import NormAdd
 from cellarium.ml.layers.transformer import Transformer, TransformerBlock
 
 __all__ = [
-    "GeneEmbedding",
-    "MetaDataEmbedding",
+    "GeneExpressionEmbedding",
+    "MetadataEmbedding",
     "MultiHeadAttention",
     "MultiHeadReadout",
     "NormAdd",
--- a/cellarium/ml/layers/attention.py
+++ b/cellarium/ml/layers/attention.py
@@ -1,0 +1,164 @@
+# Copyright Contributors to the Cellarium project.
+# SPDX-License-Identifier: BSD-3-Clause
+
+from typing import Any, Literal
+
+import torch
+from torch import nn
+from torch.nn.attention import SDPBackend, sdpa_kernel
+
+from cellarium.ml.utilities.layers import create_initializer
+
+try:
+    # use_cs returns True if the active device is a CSX device.
+    from cerebras.pytorch.backend import use_cs
+except ImportError:
+
+    def use_cs() -> bool:
+        return False
+
+
+class MultiHeadAttention(nn.Module):
+    """
+    Multi-head attention.
+
+    Args:
+        d_model:
+            Dimensionality of the embeddings and hidden states.
+        use_bias:
+            Whether to use bias in the linear transformations.
+        n_heads:
+            Number of attention heads.
+        dropout_p:
+            Dropout probability.
+        attention_logits_scale:
+            Multiplier for the attention scores.
+        attention_backend:
+            Backend for the attention computation.
+        attention_softmax_fp32:
+            Whether to use float32 for softmax computation.
+        Wqkv_initializer:
+            Initializer for the query, key, and value linear transformations.
+        Wo_initializer:
+            Initializer for the output linear transformation.
+    """
+
+    backend_map = {
+        "math": SDPBackend.MATH,
+        "flash": SDPBackend.FLASH_ATTENTION,
+        "mem_efficient": SDPBackend.EFFICIENT_ATTENTION,
+    }
+
+    def __init__(
+        self,
+        d_model: int,
+        use_bias: bool,
+        n_heads: int,
+        dropout_p: float,
+        attention_logits_scale: float,
+        attention_backend: Literal["math", "flash", "mem_efficient", "torch"],
+        attention_softmax_fp32: bool,
+        Wqkv_initializer: dict[str, Any],
+        Wo_initializer: dict[str, Any],
+    ) -> None:
+        super().__init__()
+        self.Wq = nn.Linear(d_model, d_model, bias=use_bias)
+        self.Wk = nn.Linear(d_model, d_model, bias=use_bias)
+        self.Wv = nn.Linear(d_model, d_model, bias=use_bias)
+        self.Wo = nn.Linear(d_model, d_model, bias=use_bias)
+        self.n_heads = n_heads
+        self.dropout_p = dropout_p
+        self.attention_logits_scale = attention_logits_scale
+        self.attention_backend = attention_backend
+        self.attention_softmax_fp32 = attention_softmax_fp32
+        self.Wqkv_initializer = Wqkv_initializer
+        self.Wo_initializer = Wo_initializer
+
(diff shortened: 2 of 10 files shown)


cellarium/ml/layers/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+from cellarium.ml.layers.attention import MultiHeadAttention
+from cellarium.ml.layers.embedding import GeneEmbedding, MetaDataEmbedding
+from cellarium.ml.layers.ffn import PositionWiseFFN
+from cellarium.ml.layers.head import MultiHeadReadout
+from cellarium.ml.layers.normadd import NormAdd
+from cellarium.ml.layers.transformer import Transformer, TransformerBlock
+
+__all__ = [
+    "GeneEmbedding",
+    "MetaDataEmbedding",
```
mbabadi, line 13: `MetaData` -> `Metadata`

cellarium/ml/layers/attention.py
```diff
@@ -0,0 +1,164 @@
...
+            if module.bias is not None:
+                nn.init.zeros_(module.bias)
+
+        create_initializer(self.Wo_initializer)(self.Wo.weight)
+        if self.Wo.bias is not None:
+            nn.init.zeros_(self.Wo.bias)
+
+    @staticmethod
+    def split_heads(X_nqd: torch.Tensor, n_heads: int) -> torch.Tensor:
+        """Transposition for parallel computation of multiple attention heads."""
```
mbabadi, line 91: Add `assert X_nqd.shape[-1] % n_heads == 0`, here or better yet, in the initializer of the `TransformerBlock`.

cellarium/ml/layers/attention.py
```diff
@@ -0,0 +1,164 @@
...
+    @staticmethod
+    def merge_heads(X_nhqk: torch.Tensor) -> torch.Tensor:
+        """Reverse of split_heads."""
+        X_nqhk = X_nhqk.permute(0, 2, 1, 3)
+        X_nqd = X_nqhk.reshape(X_nqhk.shape[0], X_nqhk.shape[1], -1)
+        return X_nqd
+
+    def forward(
+        self,
+        query_ncd: torch.Tensor,
```
mbabadi, line 105: `query_ncd` -> `x_query_ncd`
`key_ncd` -> `x_key_ncd`
`value_ncd` -> `x_value_ncd`

so that it is clear that these are the input embeddings.

mbabadi: (just for the inpuit args, which only affected ln 125-127 -- the rest remains the same)

cellarium/ml/layers/attention.py
```diff
@@ -0,0 +1,164 @@
...
+
+        # scale_factor is computed according to the muP paper
+        scale_factor = self.attention_logits_scale / query_nhck.shape[-1]
+
+        if (self.attention_backend == "torch") or use_cs():
+            key_nhck = key_nhck * torch.tensor(scale_factor, dtype=key_nhck.dtype)
+            attention_logits_nhcc = torch.matmul(query_nhck, key_nhck.transpose(-1, -2))
+            neg_inf = torch.tensor(float("-inf"), dtype=torch.float32)
+            attention_bias_ncc = torch.where(attention_mask_ncc, 0, neg_inf).to(attention_logits_nhcc.dtype)
+            attention_logits_nhcc += attention_bias_ncc.unsqueeze(1).expand(attention_logits_nhcc.shape)
```
mbabadi, line 141: Doesn't this work on Cerebras?
```
attention_logits_nhcc = attention_logits_nhcc.masked_fill(
    attention_mask_ncc.unsqueeze(1).expand(attention_logits_nhcc.shape) == 0, float('-inf'))
```

cellarium/ml/layers/attention.py
```diff
@@ -0,0 +1,164 @@
...
+            if self.attention_softmax_fp32 and attention_logits_nhcc.dtype != torch.float32:
+                attention_weights_nhcc = nn.functional.softmax(attention_logits_nhcc.float(), dim=-1).to(
+                    attention_logits_nhcc.dtype
+                )
+            else:
+                attention_weights_nhcc = nn.functional.softmax(attention_logits_nhcc, dim=-1)
+            attention_weights_nhcc = nn.functional.dropout(
```
mbabadi, line 148: Is it common to do dropout after softmax, or before (e.g. setting some random keys to -inf)?

cellarium/ml/layers/embedding.py
```diff
@@ -0,0 +1,103 @@
...
+
+from typing import Any
+
+import torch
+from torch import nn
+
+from cellarium.ml.utilities.layers import create_initializer
+
+
+class GeneEmbedding(nn.Module):
```
mbabadi, line 12: In principle, `GeneExpressionEmbedding` and `MetadataEmbedding` can be unified by allowing either `categorical_vocab_sizes` and `continuous_vocab_size` to be `dict[str, int] | None` (or simply observing that these dictionaries could be empty). Then you could have a `reduction_op` initializer arg, which would be `sum` or `stack`.

mbabadi: You could keep the specific classes, but the implementations would be one-liners.

cellarium/ml/layers/normadd.py
```diff
@@ -0,0 +1,48 @@
...
+
+from collections.abc import Callable
+
+import torch
+from torch import nn
+
+
+class NormAdd(nn.Module):
+    """
+    Pre-norm layer where the layer normalization is applied before the sublayer.
```
mbabadi, line 12: This is interesting... because my intuition would be to perform normalization after sublayer _and_ residual, just so that the hidden state remains normalized...

mbabadi: I guess one could imagine a scenario that the self-attention layer drives the embeddings to extreme values, and post-sublayer normalization would essentially kill everything else and only keep those extreme values. The pre-norm layer prevents this, by making sure you're adding two normalized values. Perhaps one could even entertain mixing both:
```
norm(z + sublayer(norm(z)))
```

mbabadi (changes requested):
Looks great! A few small suggestions.

cellarium/ml/layers/attention.py
ordabayevy: 1. `masked_fill` is very slow on CSX (and on GPU too I think)
2. `torch.where` is a faster alternative but I think is also a bit slow.

One worry I have is that if we use `torch.where` or `masked_fill` directly on `attention_logits_nhcc` then it has to do it for every head as well, while there is no head dimension for `attention_bias_ncc`.

But this is something that needs to be tested out.

cellarium/ml/layers/attention.py
ordabayevy: PyTorch implementation does it after the softmax - https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html

cellarium/ml/layers/normadd.py
ordabayevy: Yes, the pre-norm is the standard now.

Yang et al says

> In addition, while our results on width, batch size, sequence length, and training time
still hold for post-layernorm (Fig. 17),11 the transfer across depth only works for pre-layernorm
Transformer

> in fact, post-layernorm Transformers are much more sensitive to HPs than pre-layernorm, so our technique
is more crucial for them, especially for transfer across width. Fig. 1 uses post-layernorm"

cellarium/ml/layers/embedding.py
ordabayevy: One issue in combining them is that the inputs to the forward method have different shapes - `_nd` and `_ncd` which makes it confusing to annotate their shapes with our convention.

cellarium/ml/layers/normadd.py
ordabayevy: OLMo uses pre-LN
https://github.com/allenai/OLMo/blob/main/olmo/model.py#L671-L676

HuggingFace GPT2 use pre-LN
https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt2/modeling_gpt2.py#L614-L626

ordabayevy: Here is an interesting paper about the benefits of the LayerNorm before the attention score computation:

https://arxiv.org/pdf/2305.02582

ordabayevy: This is the original paper showing that pre-LN is more stable than  post-LN

https://proceedings.mlr.press/v119/xiong20b/xiong20b.pdf

cellarium/ml/layers/transformer.py
```diff
@@ -0,0 +1,196 @@
...
+            attention_mask_ncc:
+                Attention mask tensor of shape ``(n, c, c)``.
+
+        Returns:
+            The output hidden state tensor of shape ``(n, c, d)``.
+        """
+        for block in self.blocks:
+            hidden_state_ncd = block(hidden_state_ncd, attention_mask_ncc)
+
+        return self.ln(hidden_state_ncd)
```
ordabayevy, line 196: Added this final LN which seems to be common in pre-LN transformer architectures. I see improvements in training with it judging from the training loss curves and grad norm curves.

w/ final LN
Screenshot 2024-10-13 at 10 36 29 AM
Screenshot 2024-10-13 at 10 36 41 AM
Screenshot 2024-10-13 at 10 36 52 AM

w/o final LN
Screenshot 2024-10-13 at 10 37 27 AM
Screenshot 2024-10-13 at 10 37 34 AM
Screenshot 2024-10-13 at 10 37 43 AM

cellarium/ml/layers/transformer.py
```diff
@@ -0,0 +1,196 @@
...
+                )
+                for _ in range(n_blocks)
+            ]
+        )
+        self.ln = nn.LayerNorm(d_model, use_bias)
+
+        self._reset_parameters()
+
+    def _reset_parameters(self) -> None:
+        self.ln.reset_parameters()
```
ordabayevy, line 176: The idea behind `_reset_parameters` methods is that you can call `self.apply(lambda x: x._reset_parameters())` from the top model (e.g., `CellariumGPT`) which will recursively call `_reset_parameters` on all children.

Here `Transformer._reset_parameters` only has to reset `LayerNorm` parameters because MHA and FFN layers already have their own `_reset_parameters` methods defined.

mbabadi (approved):
Great, looks good to me!

PR #253, opened 2024-10-10
Move `cpu_transforms` back to `module_pipeline` at teardown

### LINKED ISSUE #244: Add a teardown method for `cpu_transforms` to handle multi-use case of a `CellariumModule` object
For example if you train a `CellariumModule` with `cpu_transforms` with one datamodule and then use another datamodule for prediction then `cpu_transforms` will stay attached to the training datamodule. The detaching mechanism needs to be added at the end of training/predicting.

ordabayevy:
Fixes #244

### DIFF
--- a/cellarium/ml/core/module.py
+++ b/cellarium/ml/core/module.py
@@ -10,7 +10,7 @@ import numpy as np
 import torch
 from lightning.pytorch.utilities.types import STEP_OUTPUT, OptimizerLRSchedulerConfig
 
-from cellarium.ml.core.datamodule import CellariumAnnDataDataModule, collate_fn
+from cellarium.ml.core.datamodule import CellariumAnnDataDataModule
 from cellarium.ml.core.pipeline import CellariumPipeline
 from cellarium.ml.models import CellariumModel
 from cellarium.ml.utilities.core import FunctionComposer, copy_module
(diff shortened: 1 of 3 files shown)


sjfleming:
Yes, I think this makes sense to me.  But one question, what calls the `setup()` and `teardown()` methods?

cellarium/ml/core/module.py
```diff
@@ -154,14 +154,7 @@ def configure_model(self) -> None:
...
         # move the cpu_transforms to the dataloader's collate_fn if the dataloader is going to apply them
-        if self._trainer is not None:
-            if hasattr(self.trainer, "datamodule"):
-                if isinstance(self.trainer.datamodule, CellariumAnnDataDataModule):
-                    self._cpu_transforms_in_module_pipeline = False
-                    self.trainer.datamodule.collate_fn = FunctionComposer(
-                        first_applied=collate_fn,
-                        second_applied=self.cpu_transforms,
-                    )
+        self.move_cpu_transforms_to_dataloader()
```
sjfleming, line 157: Should this be called in `configure_model ` if `setup` calls it?  I guess I'm not sure when `setup` gets called.

ordabayevy:
The way it works (my understanding) is both `setup` (first) and `configure_model` (second) get called on each device. When they are called for the first time `setup` will skip calling `move_cpu_transforms_to_dataloader`  because `pipeline` is still `None`. In this case `configure_model` is responsible for calling `move_cpu_transforms_to_dataloader`. Next time, if the module is used again with the trainer, `setup` will call `move_cpu_transforms_to_dataloader` since it has access to `pipeline` now and `configure_model` will do nothing because it should do nothing in subsequent calls.

sjfleming (approved):
Looks good

PR #256, opened 2024-10-11
Allow `AnnDataField` to have more than one key

sjfleming:
Small change that allows `key` of `AnnDataField` to be a list of strings.

An accompanying change that allows `categories_to_codes` to handle a `DataFrame` (the result of accessing `adata.obs` with a list of keys) as well as a `Series`.

This enables the scvi model to be written much more efficiently.  We have a use case where we can have several `adata.obs` keys designated as "categorical covariates", and we would like to load them all into a single tensor that gets passed to the `forward` call of the model.  These changes enable that.

### DIFF
--- a/cellarium/ml/utilities/data.py
+++ b/cellarium/ml/utilities/data.py
@@ -54,7 +54,7 @@ class AnnDataField:
     """
 
     attr: str
-    key: str | None = None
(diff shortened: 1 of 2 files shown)


ordabayevy (approved):
Cool feature! LGTM

PR #267, opened 2024-11-07
PredictWriter can choose predict() key

### LINKED ISSUE #264: Allow `PredictionWriter` to use key other than "x_ng"
Currently `PredictionWriter` is hard-coded to expect the output of model`.predict()` to be a dictionary whose value for the key `"x_ng"` is the value to be written by the prediction writer.
https://github.com/cellarium-ai/cellarium-ml/blob/ca4d6f11a4b26f6c3573a190a783066b9eaccd8f/cellarium/ml/callbacks/prediction_writer.py#L74

Might be nice to allow the key to have another name, as "x_ng" does not always make sense.  For an embedding for example, it might make more sense to call the key "z_nk" or something else.

sjfleming:
Closes #264


sjfleming:
No tests

cellarium/ml/callbacks/prediction_writer.py
```diff
@@ -56,10 +56,16 @@ class PredictionWriter(pl.callbacks.BasePredictionWriter):
             written. If not ``None``, only the first ``prediction_size`` columns will be written.
```
ordabayevy, line 56: Can you add the `key` arg to the docstring?

ordabayevy (approved):
LGTM pending the docstring fix. Let's just make @fedorgrab aware of this change.

cellarium/ml/callbacks/prediction_writer.py
sjfleming: Done

sjfleming:
Needs #282 to pass mypy tests

PR #268, opened 2024-11-08
Add `flex_attention`  backend

ordabayevy:
Removed `flash` backend because it doesn't support attention mask.


mbabadi (approved):
Great! Merge when tests pass!

PR #270, opened 2024-11-21
Re-vamp cNMF tutorial notebook

sjfleming:
Went through the notebook to get it to work in a way that can be extended to other datasets (hopefully).

Had to fix some bugs in nmf.py to do this.  Renamed `factors_kg` to `factors_rkg` for clarity, so that statements like `k = factors_rkg.shape[1]` make sense.


sjfleming:
If this PR is merged, the old tutorial notebook will be gone and the new tutorial notebook will be called `cnmf_demo.ipynb`

ImXman (approved):

PR #271, opened 2024-11-26
Automatically add lightning hooks to CellariumModule as passthroughs

sjfleming:
The idea here is to add a whole bunch of lightning hooks to CellariumModule so that CellariumModule can pass them all through to CellariumModel.  Different CellariumModels may want to use different hooks.

### DIFF
--- a/cellarium/ml/core/module.py
+++ b/cellarium/ml/core/module.py
@@ -8,7 +8,7 @@ from typing import Any
 import lightning.pytorch as pl
 import numpy as np
 import torch
-from lightning.pytorch.utilities.types import STEP_OUTPUT, OptimizerLRSchedulerConfig
+from lightning.pytorch.utilities.types import OptimizerLRSchedulerConfig
 
 from cellarium.ml.core.datamodule import CellariumAnnDataDataModule
 from cellarium.ml.core.pipeline import CellariumPipeline
@@ -329,15 +329,9 @@ class CellariumModule(pl.LightningModule):
             if callable(set_epoch):
                 set_epoch(self.current_epoch)
 
-    def on_train_start(self) -> None:
-        """
-        Calls the ``on_train_start`` method on the :attr:`model` attribute.
-        If the :attr:`model` attribute has ``on_train_start`` method defined, then
-        ``on_train_start`` must be called at the beginning of training.
-        """
-        on_train_start = getattr(self.model, "on_train_start", None)
-        if callable(on_train_start):
-            on_train_start(self.trainer)
+        on_train_epoch_start = getattr(self.model, "on_train_epoch_start", None)
+        if callable(on_train_epoch_start):
+            on_train_epoch_start(self.trainer)
 
     def on_train_epoch_end(self) -> None:
         """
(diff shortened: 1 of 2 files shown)


cellarium/ml/core/module.py
```diff
@@ -416,3 +402,63 @@ def on_save_checkpoint(self, checkpoint: dict[str, Any]) -> None:
...
                 checkpoint["loops"]["fit_loop"]["epoch_progress"]["current"]["completed"] += 1
                 checkpoint["CellariumAnnDataDataModule"]["epoch"] += 1
+
+
+# define a decorator to dynamically add methods
+def add_delegate_method(method_name):
+    def delegate_method(self, *args, **kwargs):
+        method = getattr(self.model, method_name, None)
+        if callable(method):
+            method(self.trainer)
```
ordabayevy, line 412: Pass `*args` and `**kwargs` here as well.

cellarium/ml/core/module.py
```diff
@@ -329,15 +329,9 @@ def on_train_epoch_start(self) -> None:
...
-        """
-        Calls the ``on_train_start`` method on the :attr:`model` attribute.
-        If the :attr:`model` attribute has ``on_train_start`` method defined, then
-        ``on_train_start`` must be called at the beginning of training.
-        """
-        on_train_start = getattr(self.model, "on_train_start", None)
-        if callable(on_train_start):
-            on_train_start(self.trainer)
+        on_train_epoch_start = getattr(self.model, "on_train_epoch_start", None)
+        if callable(on_train_epoch_start):
```
ordabayevy, line 333: Would be nice to somehow call the delegate_method here so if there are any changes in it this wouldn't need any changes.

cellarium/ml/core/module.py
sjfleming: So I did try this initially:
```python
method(self.trainer, *args, **kwargs)
```
but I got unexpected test failures...
then I realized that the explicitly implemented ones only used the `self.trainer` arg.

sjfleming: Should we be passing args and kwargs?

cellarium/ml/core/module.py
```diff
@@ -416,3 +402,63 @@ def on_save_checkpoint(self, checkpoint: dict[str, Any]) -> None:
...
+    setattr(CellariumModule, method_name, delegate_method)
+
+
+# list of all Lightning hook methods to delegate, which comes from:
+# >>> all_methods = dir(pl.LightningModule)
+# >>> all_hooks = [method for method in all_methods if method.startswith("on_") and not ("checkpoint" in method)]
+# >>> explicitly_implemented_hooks = ["on_train_epoch_start", "on_train_epoch_end"]
+# >>> omit_hooks = ["on_after_batch_transfer", "on_before_batch_transfer", "on_gpu"]
+# >>> passthrough_hooks = [hook for hook in all_hooks if hook not in (explicitly_implemented_hooks + omit_hooks)]
+passthrough_hooks = [
```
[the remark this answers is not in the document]
sjfleming: I was kind of thinking that as well, but I'm not sure.  On one hand, it's nice if it's automatic, but on the other hand, the code is a bit easier to read -- easier for developers to see what is actually implemented -- if it's explicit.

What do you think?

cellarium/ml/core/module.py
sjfleming: I am not sure if I understand this point.  I was thinking that we could not re-implement the explicitly implemented methods (like `on_train_epoch_start`) because the passthrough/delegate would end up having the same name and they would conflict (I don't know how python handles that).

cellarium/ml/core/module.py
sjfleming: I admit I did not understand the test failures.  It seemed like something funny was happening during "predict".  I gave up on args and kwargs for that reason, thinking it must be wrong.  But I never understood the test failure.

PR #277, opened 2024-12-11
Refactor and streamline NMF (partially)

sjfleming:
Code changes to reduce copy-and-paste and improve readability and fix bugs.


cellarium/ml/models/nmf.py
```diff
@@ -345,6 +311,32 @@ def init_weights(m):
...
+class NMFInit:
+    @staticmethod
+    def __call__(x: torch.Tensor, transformed_data_mean: float, k: int) -> None:
+        """Modify the values of x in place as a way to initialize a dictionary factor matrix for NMF."""
+        pass
+
+
+class NMFInitSklearnRandom(NMFInit):
+    @staticmethod
+    def __call__(x: torch.Tensor, transformed_data_mean: float, k: int) -> None:
```
ImXman, line 323: The online version of NMF will not see the entire data. So how can we estimated transformed data mean, if the model only see a small proportion of the entire data. If such estimation is based on a proportion of data, it can be a problematic approach.

cellarium/ml/models/nmf.py
```diff
@@ -378,163 +372,141 @@ def __init__(
...
-            optimizer.step()
-
-            alpha_diff = torch.linalg.norm(alpha_rnk.exp().view(-1, k) - alpha_buffer.view(-1, k)) / torch.linalg.norm(
-                alpha_rnk.exp().view(-1, k)
+            # update D, Mairal Algorithm 1 step 7
+            updated_factors_rkg = self._compute_factors(
+                factors_rkg=factors_rkg,
+                A_rkk=A_rkk,
+                B_rkg=B_rkg,
+                n_iterations=1000,
```
ImXman, line 506: In fitting, I usually run max 200 iterations.

cellarium/ml/models/nmf.py
```diff
@@ -378,163 +372,141 @@ def __init__(
...
+        r, k, _ = factors_rkg.shape
 
-        # updata alpha
-        alpha_rnk = torch.zeros((r, n, k), requires_grad=True, device=factors_rkg.device)
-        alpha_rnk = self.solve_alpha_wKL(alpha_rnk, x_ng, 100)
+        # update alpha, Mairal Algorithm 1 step 4
+        alpha_rnk = self._compute_loadings(
+            x_ng=x_ng,
+            factors_rkg=factors_rkg,
+            n_iterations=1000,
```
ImXman, line 489: In fitting, I usually run max 200 iterations.

ImXman (commented):
Left some comments aside the line of that code

cellarium/ml/models/nmf.py
sjfleming: Agreed, either the user would have to make a guess or precompute it.  We could additionally implement a random init that does not require this value, but we cannot then call it "SklearnRandom"

cellarium/ml/models/nmf.py
sjfleming: Tests still pass with 200, so I will change that back to 200

PR #278, opened 2024-12-12
`PyTreeDataset`

ordabayevy:


cellarium/ml/utilities/data.py
```diff
@@ -107,8 +107,6 @@ def collate_fn(
             # If so, just take the first one
             value = batch[0][key]
         elif isinstance(batch[0][key], dict):
-            if not (key.endswith("_n") or key.endswith("_ng")):
-                raise ValueError(f"Sub-dictionary '{key}' must have a batch dimension (end with '_n' or '_ng').")
```
ordabayevy, line 111: I think this is not necessary. Apart from special cases like `_g` and `_categories` the default behavior is to concatenate along the first dimension.

mbabadi (approved):
Great! Merge when tests pass!

PR #282, opened 2025-01-02
Fix issue with mypy in lightning 2.5.0.post0

### LINKED ISSUE #280: mypy error caused by lightning 2.5.0.post0
See the error here, looks like they changed something about the typing for optimizers and schedulers.
https://github.com/cellarium-ai/cellarium-ml/actions/runs/12585397381/job/35077090137

Probably all tests will now fail in github actions.

sjfleming:
Closes #280

### DIFF
--- a/cellarium/ml/core/module.py
+++ b/cellarium/ml/core/module.py
@@ -9,7 +9,7 @@ from typing import Any
 import lightning.pytorch as pl
 import numpy as np
 import torch
-from lightning.pytorch.utilities.types import STEP_OUTPUT, OptimizerLRSchedulerConfig
+from lightning.pytorch.utilities.types import STEP_OUTPUT, LRSchedulerConfigType, OptimizerLRSchedulerConfig
 
 from cellarium.ml.core.datamodule import CellariumAnnDataDataModule
 from cellarium.ml.core.pipeline import CellariumPipeline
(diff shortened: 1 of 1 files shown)


sjfleming:
I'm not sure if this is the best fix.  I've introduced a dummy scheduler which does not change the learning rate

```python
scheduler = torch.optim.lr_scheduler.ConstantLR(optimizer, factor=1.0)
```

I could not really come up with a different way, but there may be one.

Alternatively I guess we could ignore the `mypy` error.

sjfleming:
Also changed
```python
scheduler = scheduler_fn(optim_config["optimizer"], **scheduler_kwargs)
```
to
```python
scheduler = scheduler_fn(optimizer, **scheduler_kwargs)
```
but I don't know if I'm missing some reason it wasn't like that

ordabayevy:
Upon looking again at the lightning source code, it looks like the return type of `configure_optimizers` should be `OptimizerLRScheduler` which is the union of all different possible return types including `None`. With that we can return dictionaries for different cases and mypy should be able to infer the types itself:

```py
        if scheduler_fn is not None:
            scheduler = scheduler_fn(optimizer, **scheduler_kwargs)
            return {
                "optimizer": optimizer,
                "lr_scheduler": {"scheduler": scheduler, "interval": "step"},
            }
        else:
            return {"optimizer": optimizer}
```

sjfleming:
Aha, right you are.  Much better

ordabayevy (approved):
Thanks for the fix!

PR #286, opened 2025-01-17
`PredictionWriter`: optional gzip, use ThreadPoolExecutor

### LINKED ISSUE #285: PredictWriter: gzip csvs and run in background thread
Outputs from PredictWriter can be large.  It would be nice to be able to auto-gzip them as a first step.  In future, maybe we should support outputs other than CSV.

Gzipping can be handled automatically by pandas, but it takes a long time.  And the rest of the code waits.

We should have the PredictWriter acting in a background process or thread, so the main lightning code can continue without waiting.

This can take an scvi data reconstruction run from 18 hours to 5 hours (preliminary testing of a fix using `ThreadPoolExecutor`).

sjfleming:
Closes #285 

These changes add two init args to `PredictionWriter`: `gzip` (bool) and `max_threadpool_workers` (int), each of which have default values.

`PredictionWriter` now gzips saved CSVs by default, and runs the saving-and-gzip process in a background thread that does not block further `lightning` compute.

`NewLightningCLI` in `cli.py` also now injects `return_predictions=False` into calls to `trainer.predict()`.

Testing indicates the following outcomes for scvi reconstructions which involve computing a dense output CSV with 250 columns:

```
               |    before changes     |     after changes
---------------------------------------------------------------
file size      |            12 MB      |        5 MB
total time     |            18 hr      |        5 hr
```

(the 18 hrs is gzipping the CSVs without the ThreadPool.  if you don't gzip, the total time would be 7.5 hours.)

### DIFF
--- a/cellarium/ml/cli.py
+++ b/cellarium/ml/cli.py
@@ -291,6 +291,19 @@ def lightning_cli_factory(
             # https://github.com/Lightning-AI/pytorch-lightning/pull/18105
             pass
 
+        def before_instantiate_classes(self):
+            # issue a UserWarning if the subcommand is predict and return_predictions is not set to False
+            if self.subcommand == "predict":
+                return_predictions: bool = self.config["predict"]["return_predictions"]
+                if return_predictions:
+                    warnings.warn(
+                        "The `return_predictions` argument should be set to 'false' when running predict to avoid OOM. "
+                        "This can be set at indent level 0 in the config file. Example:\n"
+                        "model:  ...\ndata:  ...\ntrainer:  ...\nreturn_predictions: false",
+                        UserWarning,
+                    )
+            return super().before_instantiate_classes()
+
         def instantiate_classes(self) -> None:
             with torch.device("meta"):
                 # skip the initialization of model parameters
@@ -624,21 +637,6 @@ def main(args: ArgsType = None) -> None:
     if model_name not in REGISTERED_MODELS:
         raise ValueError(f"'model_name' must be one of {list(REGISTERED_MODELS.keys())}. Got '{model_name}'")
     model_cli = REGISTERED_MODELS[model_name]
-
-    # issue a UserWarning if the subcommand is predict and return_predictions is not set to False
-    if isinstance(args, list):
-        subcommand = args[0]
-        return_predictions = None  # will always issue UserWarning
-    else:
-        subcommand = args.get("subcommand", None)
-        return_predictions = args.get("predict", {}).get("return_predictions", None)
-    if (subcommand == "predict") and (return_predictions != "false"):
-        warnings.warn(
-            "The `return_predictions` argument should be set to 'false' when running predict to avoid OOM. "
-            "This can be set at indent level 0 in the config file. Example:\n"
-            "model:  ...\ndata:  ...\ntrainer:  ...\nreturn_predictions: false",
-            UserWarning,
-        )
 
     with warnings.catch_warnings():
         warnings.filterwarnings("ignore", message="Transforming to str index.")
--- a/cellarium/ml/cli.py
+++ b/cellarium/ml/cli.py
@@ -309,6 +309,13 @@ def lightning_cli_factory(
                 }
             )
 
+        def predict(self, *args, **kwargs):
+            """Not well documented, but defining this here overrides the default predict subcommand.
+            This method injects return_predictions=False into the kwargs to prevent the predictions from
+            being returned, which prevents memory overflow when writing predictions to a file."""
+            kwargs["return_predictions"] = False
+            self.trainer.predict(*args, **kwargs)
+
     return NewLightningCLI
 
 
--- a/cellarium/ml/cli.py
+++ b/cellarium/ml/cli.py
@@ -309,13 +309,6 @@ def lightning_cli_factory(
                 }
             )
 
-        def predict(self, *args, **kwargs):
-            """Not well documented, but defining this here overrides the default predict subcommand.
-            This method injects return_predictions=False into the kwargs to prevent the predictions from
-            being returned, which prevents memory overflow when writing predictions to a file."""
-            kwargs["return_predictions"] = False
-            self.trainer.predict(*args, **kwargs)
-
     return NewLightningCLI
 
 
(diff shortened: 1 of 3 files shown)


sjfleming:
The only thing undesirable about this is that the `ThreadPoolExecutor` can fall behind.  In my current test run for example, the prediction writing seems to be about 100 batches behind the actual lightning compute.  There might be a danger of OOM if the `max_threadpool_workers` is so low that you're really lagging behind in terms of writing the outputs.

sjfleming:
Empirically it seems that 8 thread workers does not fall behind systematically, so for now I will change the default to 8

sjfleming:
Ah dang it, it was killed with 8 workers due to OOM after 2.5 hours...
that may be because I pushed batch size too high, but perhaps this approach is a bit fragile

sjfleming:
Okay I have now implemented a `BoundedThreadPoolExecutor` which prevents the executor's queue from growing without limit (and thus memory usage growing without limit).  Let's see if this works.

Projected total runtime looks like about 6 hours now instead of 5.

sjfleming:
Added an (untested) check to fail fast if it can be projected that the total size of the prediction output files will not fit in the allocated disk space.  (I ran into this problem and it was only after several hours I found out.  :(  )

sjfleming:
Needed to reach into `cli.py` to implement the note in the docstring here:
https://github.com/cellarium-ai/cellarium-ml/blob/090161e2baa1857d07850c7c18fea8dd09a056ef/cellarium/ml/callbacks/prediction_writer.py#L47-L49

i.e. the `return_predictions=False` kwarg needs to be passed to `trainer.predict()`.  I think the changes to `cli.py` are the correct way to make this happen.

sjfleming:
The above `cli.py` modification did fix the full scvi run.

cellarium/ml/cli.py
```diff
@@ -309,6 +309,13 @@ def add_arguments_to_parser(self, parser: LightningArgumentParser) -> None:
                 }
             )
 
+        def predict(self, *args, **kwargs):
+            """Not well documented, but defining this here overrides the default predict subcommand.
+            This method injects return_predictions=False into the kwargs to prevent the predictions from
+            being returned, which prevents memory overflow when writing predictions to a file."""
+            kwargs["return_predictions"] = False
+            self.trainer.predict(*args, **kwargs)
+
```
ordabayevy, line 318: I think `return_predictions` can already be passed to the config file in default LightningCLI:

https://github.com/cellarium-ai/cellarium-ml/blob/090161e2baa1857d07850c7c18fea8dd09a056ef/tests/test_cli.py#L399

ordabayevy: I think the note about the `return_predictions` needs to be fixed because it is passed not to the init method (as it is implied by the note) but to the predict method via an additional argument in the config file:

```yaml
model:
  ...
data:
  ...
trainer:
  ...
return_predictions: false
```

sjfleming: Aw man, okay, that's what I was missing

sjfleming: I was trying to pass it from the config file, but had no idea where to put it and guessed all the wrong places.  How did you figure out that it goes there?  I spent quite some time reading docs... but never figured it out...

sjfleming:
I can get rid of the changes to `cli.py` certainly, and leave `return_predictions: false` to the config file if you think that makes the most sense.

The only thing that bothers me about that solution is that somebody like me can come in, not know they have to include that in the config file or where it goes, and waste a bunch of time hitting out-of-memory errors   :)

What I like about modifying `cli.py` in the manner above is that `return_predictions` becomes `False` whenever `predict` is called, regardless of the config file.  So it prevents users from making mistakes.  My thinking is just that the simpler the config files can be, the better.

But I can see both sides... and I guess it would be okay with me if we left things as is and just put in some kind of massive `UserWarning` or something to check and make sure someone hasn't forgotten to add `return_predictions: false` to their config file by mistake.

What do you think @ordabayevy ?

cellarium/ml/cli.py
ordabayevy: It was the opposite experience for me. I initially generated a default predict config file with `... predict --print_config` which contained `return_predictions` and then had to figure out what it does from the documentation :)

ordabayevy:
I prefer using the config file instead of hard coding and do the following:

1. Improve the documentation of `PredictionWriter` and add an example of how `return_predictions` should be added to the config file.
2. If you think that is not enough, add a warning message if `return_predictions=True`.

... but if we never really gonna need `return_predictions=True` maybe it makes sense to hard code it, idk. What happens if both your changes applied and `return_predictions=True` is set?

sjfleming:
Okay I've got those changes implemented.  Got rid of the hard-coded `return_predictions=False`.  Got rid of fail-fast if predictions won't fit on disk (thinking ahead to #290 ).  Issues a `UserWarning` if running `predict` with `return_predictions=True`.  Included explicit config file example in the docstring for `PredictionWriter`.

cellarium/ml/cli.py
```diff
@@ -624,6 +624,22 @@ def main(args: ArgsType = None) -> None:
...
+    else:
+        subcommand = args.get("subcommand", None)
+        return_predictions = args.get("predict", {}).get("return_predictions", None)
+    if (subcommand == "predict") and (return_predictions != "false"):
+        warnings.warn(
+            "The `return_predictions` argument should be set to 'false' when running predict to avoid OOM. "
+            "This can be set at indent level 0 in the config file. Example:\n"
+            "model:  ...\ndata:  ...\ntrainer:  ...\nreturn_predictions: false",
+            UserWarning,
+        )
```
ordabayevy, line 641: I think this might not work if the config file used. Config file is parsed later on in the init method of `LightningCLI`. Probably this  logic needs to be moved somewhere there. It should happen after this line [         self.parse_arguments(self.parser, args)](https://github.com/Lightning-AI/pytorch-lightning/blob/a944e7744e57a5a2c13f3c73b9735edf2f71e329/src/lightning/pytorch/cli.py#L384). This hook https://github.com/Lightning-AI/pytorch-lightning/blob/a944e7744e57a5a2c13f3c73b9735edf2f71e329/src/lightning/pytorch/cli.py#L554 might be a good place.

tests/test_cli.py
```diff
@@ -631,3 +633,135 @@ def test_compute_var_names_g(tmp_path: Path) -> None:
...
+    with pytest.warns(UserWarning, match=match_str) as record:
+        warnings.warn("we need one warning: " + match_str, UserWarning)
+        main(copy.deepcopy(config))
+    n = 0
+    for r in record:
+        assert isinstance(r.message, Warning)
+        warning_message = r.message.args[0]
+        if match_str in warning_message:
+            n += 1
+    assert n < 2, "Unexpected UserWarning when running predict with return_predictions=false"
```
ordabayevy, line 703: What does this test do?

ordabayevy (approved):
Looks great!

tests/test_cli.py
sjfleming: Yeah so this is asserting that the UserWarning is _not_ emitted if running prediction with `return_predictions: false`.

I might be doing it in a weird way, but I'm not sure what the right way is.  It's easy to assert that a warning is emitted, but not so easy to test that a warning is _not_ emitted (from what I can tell).  The only way I could figure was to count up the warnings matching a certain match string.  (And I needed at least one such warning, or the counting mechanism would not work.  Thus the assertion `n < 2`... there is one "fake" warning to enable counting, and then any further warning would be the real warning.)

PR #289, opened 2025-01-21
Fix the logic for merging odd number of devices in IncrementalPCA

ordabayevy:
This fixes a bug in IncrementalPCA. There is a logic that checks that an even rank should be merged with an odd rank only when a corresponding odd rank exists. Instead of checking that even rank is less than the world size `rank < world_size` it should check that the corresponding odd rank is less than the world size `rank + 1 < world_size`. Without this fix the model just halts because it cannot receive the tensor from another rank.

To include it in the test, I added `TEST_DEVICES=3` to the test matrix. This requires fixing some tests by making the number of cells divisible by three so that no cells are dropped.

### DIFF
--- a/cellarium/ml/models/incremental_pca.py
+++ b/cellarium/ml/models/incremental_pca.py
@@ -89,15 +89,11 @@ class IncrementalPCA(CellariumModel, PredictMixin):
         assert_columns_and_array_lengths_equal("x_ng", x_ng, "var_names_g", var_names_g)
         assert_arrays_equal("var_names_g", var_names_g, "var_names_g", self.var_names_g)
 
-        g = self.n_vars
         k = self.n_components
         niter = self.svd_lowrank_niter
         self_X_size = self.x_size
         other_X_size = x_ng.size(0)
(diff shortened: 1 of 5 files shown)


cellarium/ml/models/incremental_pca.py
```diff
@@ -89,15 +89,11 @@ def forward(self, x_ng: torch.Tensor, var_names_g: np.ndarray) -> dict[str, torc
...
 
-        g = self.n_vars
         k = self.n_components
         niter = self.svd_lowrank_niter
         self_X_size = self.x_size
         other_X_size = x_ng.size(0)
         total_X_size = self_X_size + other_X_size
-        assert k <= min(other_X_size, g), (
-            f"Rank of svd_lowrank (n_components): {k} must be less than min(n_obs, n_vars): {min(other_X_size, g)}"
-        )
```
ordabayevy, line 100: I think this assertion is actually not necessary.

mbabadi (approved):
Good catch!!

PR #292, opened 2025-01-23
Address flaky mup test

### LINKED ISSUE #291: Tests: flakiness in a mup test
I've noticed that there is a mup-related test that fails at a low rate at random.  Re-running can make the test run succeed.  This is an example of the error:

https://github.com/cellarium-ai/cellarium-ml/actions/runs/12933664691/job/36072972217

The issue seems to be something related to URL requests timing out

```pytb
urllib.error.URLError: <urlopen error [Errno 110] Connection timed out>
```

This line in the test suite seems to be causing the error:
https://github.com/cellarium-ai/cellarium-ml/blob/090161e2baa1857d07850c7c18fea8dd09a056ef/tests/test_mup.py#L130

I wonder if we could retry on failure?

sjfleming:
Closes #291 

Introduces a test dependency on the `tenacity` package.

Automatically retries the download of CIFAR10 if there is a urllib connection timeout error.  Currently this implements 3 tries with 10 seconds between each try.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,7 @@ lint = ["ruff"]
 mypy = ["mypy", "types-PyYAML"]
 test = [
   "pytest-xdist",
+  "tenacity",
   "tensorboard",
   "torchvision",
 ]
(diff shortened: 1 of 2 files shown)


ordabayevy (approved):
Thanks for fixing! Hopefully it won't be an issue anymore.

PR #298, opened 2025-02-03
Fix recent mypy errors

sjfleming:
Closes #297

PR #299, opened 2025-02-04
Add to gitignore

sjfleming:
Suggestion: add a couple file types to gitignore, things I frequently end up with in the repo, like output csv files, input tsv files, h5ad data files, and checkpoint files for trained models.


sjfleming:
This needs #298 to pass mypy tests

PR #300, opened 2025-02-05
Refactor and expand onepass model

### LINKED ISSUE #163: Implement Welford's online algorithm for variance calculation
Here: https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance

It is more stable than both the naive and shifted.

sjfleming:
Closes #163 

This is a refactor of onepass to make it more extensible.  It implements the Welford algorithm for online variance calculation and it also implements a gene-gene covariance computation via an online algorithm similar to Welford (https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Covariance subsection "Online").  The latter can be run using the raw data or ranks (which allows for a computation of gene-gene Spearman correlations).

Currently the Welford implementation is actually implemented in a different cellarium class, as is the covariance implementation.  I thought this might be cleaner than having one huge class with more input arguments, but I'm open to opposing views.

Todo:

- [ ] the refactor of the existing naive and shifted algorithms must pass existing tests
- [ ] add tests for the Welford variance calculation
- [ ] add tests for the Welford-like covariance calculation
- [ ] add tests for the Welford-like correlation calculation
- [ ] add tests for the Welford-like Spearman correlation calculation

### DIFF
--- a/cellarium/ml/models/__init__.py
+++ b/cellarium/ml/models/__init__.py
@@ -5,7 +5,7 @@ from cellarium.ml.models.geneformer import Geneformer
 from cellarium.ml.models.incremental_pca import IncrementalPCA
 from cellarium.ml.models.logistic_regression import LogisticRegression
 from cellarium.ml.models.model import CellariumModel, PredictMixin, ValidateMixin
-from cellarium.ml.models.onepass_mean_var_std import OnePassMeanVarStd
+from cellarium.ml.models.onepass_gene_stats import NaiveOnlineGeneStats as OnePassMeanVarStd
 from cellarium.ml.models.probabilistic_pca import ProbabilisticPCA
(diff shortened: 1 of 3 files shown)


sjfleming:
Welford covariance computation on all 33k genes spikes to like 28GB memory usage on my laptop (eyeballing Activity Monitor).  Just wanted to make a note of this as a ballpark figure.  I am guessing that's why the github actions runner went OOM on [5495ce0](https://github.com/cellarium-ai/cellarium-ml/pull/300/commits/5495ce08d73b31d2c2f7d74f6f37b80ab020c54f).  The CLI test now uses a `Filter` and only computes covariance using a handful of genes.

sjfleming:
Ensure batch is not empty in `update()`

PR #301, opened 2025-02-06
Add FSDP support

ordabayevy:
- Handle gradient clipping by norm
- Handle sharded param and grad norm computation


mbabadi (approved):
Cool :D

PR #304, opened 2025-02-25
Add `CellariumModule.test_step` method for any custom inference

ordabayevy:
By adding `test` method to your model you can perform any kind of custom inference by running `cellarium-ml your_model test --config config_path`


sjfleming (approved):
Looks good to me!

PR #307, opened 2025-04-03
NMF loading amortization

### LINKED ISSUE #306: nmf: amortize the inference of the cell loadings, alpha
The inference of the cell loadings can be amortized by using an encoder neural network.

This may have several advantages:
- speed
- the fact that cells with similar expression profiles should have similar loadings
- you end up with an encoder so that "solving" for the loadings is quite easy at the end

sjfleming:
Closes #306 

Implements the `"amortized_mairal"` variant of the Mairal algorithm, where the loadings alpha are computed using a shared encoder whose parameters are learned.

### DIFF
--- a/cellarium/ml/models/nmf.py
+++ b/cellarium/ml/models/nmf.py
@@ -223,6 +223,8 @@ def compute_loadings(
 
     for i in range(n_iterations):
         optimizer.zero_grad()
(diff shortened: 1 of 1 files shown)


sjfleming:
Closed in favor of #362

PR #318, opened 2025-06-27
Add KL weight annealing to scvi

### LINKED ISSUE #316: scvi: enable KL annealing
Enable a cellarium model (in a way that plays nicely with lightning) to anneal some prefactor in a loss function -- in this case, a prefactor for the KL divergence in the loss.

It should be able to follow some kind of a schedule during training (like a linear ramp-up to 1 in our case, over a defined number of epochs).

Figure out the RIGHT way to do this in lightning.  Look for examples.

sjfleming:
Closes #316 

Add KL annealing to the scvi model.

This hard-codes the addition of "epoch" to `batch`, which we probably don't want to do for all models all the time.  But for scvi, we'd like to run without specifying `"test_mode"`.  But dadc "test_mode" already attaches "epoch" to the batch.

### DIFF
--- a/cellarium/ml/models/scvi.py
+++ b/cellarium/ml/models/scvi.py
@@ -7,13 +7,13 @@ import importlib
 import itertools
 from typing import Any, Literal, Sequence
 
+import lightning.pytorch as pl
 import numpy as np
 import pandas as pd
 import torch
 from anndata import AnnData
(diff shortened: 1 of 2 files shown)


sjfleming:
Do not merge yet... there seems to be a regression introduced here

PR #325, opened 2025-07-28
python 3.12 upgrade, and backed mode for h5ad reading

### LINKED ISSUE #324: Data loaders should read anndata files from disk in backed mode
Have all dataloaders load h5ad files from disk in ``backed="r"` mode.  This can probably be implemented in a few lines here:
https://github.com/cellarium-ai/cellarium-ml/blob/main/cellarium/ml/data/fileio.py

Make sure everything still works!

### LINKED ISSUE #400: lost three-device tests
In #135 somehow I lost the 3-device test case for the regular (model and core) unit tests.  Bring it back.

sjfleming:
Closes #324 

Closes #400 

A major other change is that it upgrades the whole codebase to python 3.12 and allows `lightning` versions `>2.5.2` (previously these were explicitly disallowed).  So it'll help in the long run.  It may affect checkpoint compatibility.  It should be okay, but it has to set checkpoint weights_only loading to False.

The upgrade to python 3.12 is actually to get the newer versions of anndata, which have an explicit python 3.12 requirement. The anndata release that has the backed mode fixes relies on python 3.12.


sjfleming:
@HugoHakem I got some help from copilot in creating a way to make it seem like we have a massive h5ad file that is fake.  The thing actually does not take up a lot of disk space, but if anndata tries to read it into memory, it will allocate 40GB memory and crash.
```python
@pytest.fixture
def massive_h5ad(tmp_path: Path) -> Path:
    import h5py
    import numpy as np
    
    # Create a dataset that CLAIMS to be ~40GB but uses almost no disk space
    n_obs = 2_000_000  # 2 million cells  
    n_vars = 5_000     # 5k genes
    
    h5ad_path = tmp_path / "massive_fake.h5ad"
    
    with h5py.File(h5ad_path, "w") as f:
        # Create X dataset with claimed huge size but minimal actual storage
        # Using fillvalue=0.0 with chunking - chunks are only allocated when written to
        f.create_dataset(
            "X", 
            shape=(n_obs, n_vars),
            dtype=np.float32,
            fillvalue=0.0,
            chunks=True,  # Enable chunking so not all data needs to be stored
            compression=None  # No compression to keep it simple
        )
        
        # Create minimal obs metadata - just the index is required
        obs_group = f.create_group("obs")
        # Create a small obs index but tell HDF5 it could expand to n_obs
        obs_index_data = np.array([f"CELL_{i:07d}".encode('utf-8') for i in range(min(1000, n_obs))])
        obs_group.create_dataset("_index", data=obs_index_data, maxshape=(n_obs,), dtype="S12")
        
        # Create minimal var metadata - just the index is required  
        var_group = f.create_group("var")
        var_index_data = np.array([f"GENE_{i:05d}".encode('utf-8') for i in range(n_vars)])
        var_group.create_dataset("_index", data=var_index_data, dtype="S10")
        
        # Set minimal h5ad format attributes that anndata expects
        f.attrs["encoding-type"] = "anndata"
        f.attrs["encoding-version"] = "0.1.0"
        
    return h5ad_path
```

Activity monitor when I run a pytest test that tries to read this thing in `backed=False` mode:
image

sjfleming:
It's kinda cool as a check... but it might not be great as a real pytest test because you don't really see a clear failure message.  It'll just crash your computer if it doesn't work.

HugoHakem:
Ooh super cool, good catch on that!
> @HugoHakem I got some help from copilot in creating a way to make it seem like we have a massive h5ad file that is fake. The thing actually does not take up a lot of disk space, but if anndata tries to read it into memory, it will allocate 40GB memory and crash.
> 
> ```python
> ...
> ```

So we could use this strategy into the other test to see what happens for large anndata in backed mode (Like can it be processed or not).

cellarium/ml/data/distributed_anndata.py
```diff
@@ -9,7 +9,13 @@
 import pandas as pd
 from anndata import AnnData, concat
 from anndata._core.index import _normalize_indices
-from anndata.compat import Index, Index1D
+
+try:
+    from anndata.compat import Index, Index1D
+except ImportError:
+    from anndata.typing import Index, Index1D
```
fedorgrab, line 16: When does this conditioning happen?

fedorgrab (approved):
I left just one question.:)

cellarium/ml/data/distributed_anndata.py
sjfleming: Ah yeah, this is for different anndata versions.  i think one is needed for anndata < 0.13 and one is for >

sjfleming:
TODO

- [x] change default backed mode to False
- [x] verify that streaming files from bucket is not broken

Testing on a single T4 indicates that `backed_mode='r'` is slower: nearly twice as slow when training on 10k h5ad shards.

sjfleming:
Verified that scVI and socam both train with existing config files (from before this PR) on a T4 gpu

sjfleming:
Fixed a bug in the GCS file loader

PR #328, opened 2025-08-01
scvi: implement KL annealing, tons of tests, and fix a bug

sjfleming:
Important things:

- There was a bug where, during "predict()" and "reconstruct()", the value that was being used for `z_nk` was a _sample_ from the z distribution rather than the mean. This was a really big problem leading to bad outcomes when computing latent space metrics.
- Implemented KL annealing, which is really important for scvi-tools to get as nice of results as it gets
- Added tons of tests to show parity between cellarium and scvi-tools
    - losses and gradients now match exactly! (when computed using manual forward passes)
    - allowing the codebases themselves to train the respective models, they are on par when computing nearest-neighbors in the latent space using held-out test data

### DIFF
--- a/cellarium/ml/models/scvi.py
+++ b/cellarium/ml/models/scvi.py
@@ -912,6 +912,10 @@ class SingleCellVariationalInference(CellariumModel, PredictMixin):
             "z_nk": inference_outputs["z"],
         }
 
+    def _latent_value_from_latent_distribution(self, d: Distribution) -> torch.Tensor:
+        """Get the latent variable from the latent distribution."""
+        return d.mean
+
     def predict(
         self,
         x_ng: torch.Tensor,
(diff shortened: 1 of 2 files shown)


LysSanzMoreta (commented):
@sjfleming 4 easy comments

a) we do not have "lr_factor" implemented or so?
b)  think "kl_warmup_epochs" should not be allowed to be None, just int (like in scanpy). Like that when kl_warmup_epochs == 0, then max_epochs >= kl-warmup_epochs and kl_annealing_weight = 1. Otherwise it is confusing, I think.

c) I have fixed these errors in my branch (I will push it):
      1) self.reconstruction_genes = reconstruction_genes is missing
      2) self.build_reconstructions = build_reconstructions is missing
      3) compute_annealed_kl_weight is called twice, one should be removed

sjfleming:
I'm going to go ahead and merge this now.

PR #330, opened 2025-08-06
mypy update fix

sjfleming:
Just ignoring this complaint from `mypy` caused by a pytorch update:

```python
dist.Normal(self.mean_g + z_nk @ self.W_kg, self.sigma).to_event(1)
```
gives
```pytb
cellarium/ml/models/probabilistic_pca.py:137: error: Argument 2 to "Normal" has incompatible type "PyroParam"; expected "Tensor | float"  [arg-type]
```


sjfleming:
closed in favor of #342

PR #342, opened 2025-08-26
Small mypy fix

sjfleming:
Small ignore due to a torch update


sambatsy (approved):
pypi test went successful

PR #352, opened 2025-10-16
mypy fix for new torch 2.9.0

sjfleming:
Fixes

```pytb
mypy cellarium tests
cellarium/ml/models/probabilistic_pca.py:147: error: Argument 1 has incompatible type "PyroParam"; expected "Tensor | int | float | bool | complex"  [arg-type]
Found 1 error in 1 file (checked 83 source files)
make: *** [typecheck] Error 1
```

by ignoring it

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
   "crick>=0.0.4",
   "google-cloud-storage",
   "jsonargparse[signatures]==4.27.7",
-  "lightning>=2.2.0",
+  "lightning>=2.2.0,<=2.5.2",
   "numpy<=1.26.4",
   "pyro-ppl>=1.9.1",
   "pytest",


sjfleming:
Another error... the `lightning` update to 2.5.5 now throws this error when the CLI loads a saved checkpoint:

```pytb
WeightsUnpickler error: Unsupported global: GLOBAL cellarium.ml.core.module.CellariumModule was not an allowed global by default. Please use `torch.serialization.add_safe_globals([cellarium.ml.core.module.CellariumModule])` or the `torch.serialization.safe_globals([cellarium.ml.core.module.CellariumModule])` context manager to allowlist this global if you trust this class/function.
```

```pytb
_pickle.UnpicklingError: Weights only load failed. This file can still be loaded, to do so you have two options, do those steps only if you trust the source of the checkpoint. 
        (1) In PyTorch 2.6, we changed the default value of the `weights_only` argument in `torch.load` from `False` to `True`. Re-running `torch.load` with `weights_only` set to `False` will likely succeed, but it can result in arbitrary code execution. Do it only if you got the file from a trusted source.
        (2) Alternatively, to load with `weights_only=True` please check the recommended steps in the following error message.
        WeightsUnpickler error: Unsupported global: GLOBAL cellarium.ml.core.module.CellariumModule was not an allowed global by default. Please use `torch.serialization.add_safe_globals([cellarium.ml.core.module.CellariumModule])` or the `torch.serialization.safe_globals([cellarium.ml.core.module.CellariumModule])` context manager to allowlist this global if you trust this class/function.
```

Not sure yet how to fix this in a nice way.

PR #354, opened 2025-10-20
SOCAM integration

fedorgrab:


sjfleming:
Let's add a test for socam and then also just make sure all the CI tests pass

`make format`
`make typecheck`
will help

sjfleming:
Todo: @magrenimish take a look at the unnormalized categorical dist.  Is the code implementation really unnormalized as intended?

sjfleming:
For now, let's leave the categorical implementation as it is.

BUT... here's my question @magrenimish : if we leave the categorical like it is written now (normalized), do we need to implement it ourselves?  Or can we just use Pyro's categorical?

EDIT [later]: it is fine the way it is written in the codebase currently. It does serve a function.

sjfleming:
Looks like we forgot the `Transform` called `EncodedTargets`.  I've added it back in.

sjfleming:
Add socam test to test_cli

cellarium/ml/models/socam.py
```diff
@@ -0,0 +1,333 @@
...
+
+from cellarium.ml.distributions import PyroCategorical
+from cellarium.ml.models.model import CellariumModel, PredictMixin, ValidateMixin
+from cellarium.ml.utilities.testing import (
+    assert_arrays_equal,
+    assert_columns_and_array_lengths_equal,
+)
+
+
+def compute_valid_mask(input_categories: list[str], output_categories: list[str]) -> list[int]:
```
magrenimish: this function is unused in this script . The _get_subset_info function seems to be sufficient.

tests/test_socam.py
magrenimish, line 1: tests the compute_valid_mask function from the socam script which is unused.

cellarium/ml/distributions/__init__.py
```diff
@@ -2,5 +2,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 from cellarium.ml.distributions.negative_binomial import NegativeBinomial
+from cellarium.ml.distributions.unnormalized_categorical import PyroCategorical
 
-__all__ = ["NegativeBinomial"]
+__all__ = ["NegativeBinomial", "PyroCategorical"]
```
magrenimish: should this be renamed to "UnnormaizedPyroCategorical" to avoid confusion with regular PyroCategorical?

cellarium/ml/models/socam.py
sjfleming: excellent thanks!

cellarium/ml/distributions/__init__.py
sjfleming: Yes, I'll call it `UnconstrainedPyroCategorical` I think

tests/test_socam.py
sjfleming: will get rid of that test

sjfleming:
Actually I think the way I'm handling subsetting of the ontology does not make sense.  I've been taking all cells annotated in the dataset as my subset.  But what if some ancestor nodes from the ontology are missing?  This will mess up probability propagation I think!

sjfleming:
Sparse dataloader strategy #387 takes SOCAM train time (on 1001 gzip-compressed persistent-disk local h5ad files) per epoch on a T4 gpu from 20 mins to 5 mins.

fedorgrab:
> Sparse dataloader strategy #387 takes SOCAM train time (on 1001 gzip-compressed persistent-disk local h5ad files) per epoch on a T4 gpu from 20 mins to 5 mins.

That's incredible!

sjfleming (approved):
Looks good now!

PR #370, opened 2026-04-09
Make `DivideByScale` and `ZScore` more flexible

### LINKED ISSUE #225: Make transforms adaptable to the input
Currently, transforms that rely on `var_names_g` (`ZScore` and `DivideByScale`) throw an error if input `var_names_g` doesn't match `self.var_names_g`. For example, if the `Filter` transform is used upstream then  these transforms need to be instantiated with filtered `__init__` arguments which makes it inconvenient to use. This issue proposes to make this transforms flexible and adapt to filtered or re-ordered inputs.

sjfleming:
Closes #225 
Closes #357

### DIFF
--- a/cellarium/ml/transforms/divide_by_scale.py
+++ b/cellarium/ml/transforms/divide_by_scale.py
@@ -13,7 +13,7 @@ from cellarium.ml.utilities.testing import (
 )
 
 
-class DivideByScale(nn.Module, FilterCompatibilityMixin):
+class DivideByScale(FilterCompatibilityMixin, nn.Module):
     """
     Divide gene counts by a scale.
 
--- a/cellarium/ml/transforms/divide_by_scale.py
+++ b/cellarium/ml/transforms/divide_by_scale.py
@@ -2,20 +2,18 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 
-from functools import cache
-
 import numpy as np
 import torch
 from torch import nn
 
-from cellarium.ml.utilities.data import get_var_names_g_indices
+from cellarium.ml.transforms.mixins import FilterCompatibilityMixin
 from cellarium.ml.utilities.testing import (
     assert_columns_and_array_lengths_equal,
     assert_nonnegative,
 )
 
 
-class DivideByScale(nn.Module):
+class DivideByScale(nn.Module, FilterCompatibilityMixin):
     """
     Divide gene counts by a scale.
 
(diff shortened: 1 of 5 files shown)


sjfleming:
`DivideByScale` and `ZScore` can now be initialized with all genes (like from a `OnePassMeanVarStd` run with no `Filter`), and then they can be used in a pipeline either before _or after_ a `Filter`.  This makes them much easier to use, since you don't have to worry so much about the order of operations.

cellarium/ml/transforms/divide_by_scale.py
```diff
@@ -38,6 +40,10 @@ def __init__(self, scale_g: torch.Tensor, var_names_g: np.ndarray, eps: float =
         assert_nonnegative("eps", eps)
         self.eps = eps
 
+    @cache
+    def _get_indices(self, var_names_g: tuple) -> np.ndarray:
+        return get_var_names_g_indices(np.array(var_names_g), self.var_names_g)
```
fedorgrab: It's a very tiny design suggestion. Not even sure if it's worth it, it will pay off if we will have more transforms that need that _get_indices tho. Just leaving this idea here if you like it:

Factor out this logic to a mixin class (couldn't think of a good name for this mixin class).
```python3
class VarNameIndexableMixin:
    var_names_g: np.ndarray 

    @cache
    def _get_indices(self, var_names_g: tuple) -> np.ndarray:
        return get_var_names_g_indices(np.array(var_names_g), self.var_names_g)
```

and then use it in ZScore and DivideBy Scale.

fedorgrab (approved):
Looks good, left one comment.

cellarium/ml/transforms/divide_by_scale.py
sjfleming: cool idea, I'll do it

cellarium/ml/transforms/divide_by_scale.py
```diff
@@ -6,14 +6,14 @@
...
+from cellarium.ml.transforms.mixins import FilterCompatibilityMixin
 from cellarium.ml.utilities.testing import (
-    assert_arrays_equal,
     assert_columns_and_array_lengths_equal,
     assert_nonnegative,
 )
 
 
-class DivideByScale(nn.Module):
+class DivideByScale(nn.Module, FilterCompatibilityMixin):
```
fedorgrab: Small nit: it’s a common pattern to put mixins before the base class (MRO is left → right), so their methods take priority.
Doesn’t change behavior here, but good to keep things consistent.

fedorgrab (commented):
left another comment

PR #372, opened 2026-04-11
Output csv for `OnePassMeanVarStd` and `FileMultiLoader` for configs

sjfleming:
The idea here was to have the onepass model output a simple CSV summary file, so downstream we never really need to load a checkpoint file unless we want to.

I make use of that CSV file in the `test-examples` pipeline.  Now the PCA training uses that output CSV from onepass (instead of the checkpoint).

This also seemed like a good opportunity to introduce a `FileMultiLoader` for the config files.  Instead of configs that look like this
```yaml
    - class_path: cellarium.ml.transforms.ZScore
      init_args:
        mean_g:
          !CheckpointLoader
          file_path: /tmp/test_examples/onepass/lightning_logs/version_0/checkpoints/epoch=0-step=2.ckpt
          attr: model.mean_g
          convert_fn: null
        std_g:
          !CheckpointLoader
          file_path: /tmp/test_examples/onepass/lightning_logs/version_0/checkpoints/epoch=0-step=2.ckpt
          attr: model.std_g
          convert_fn: null
        var_names_g:
          !CheckpointLoader
          file_path: /tmp/test_examples/onepass/lightning_logs/version_0/checkpoints/epoch=0-step=2.ckpt
          attr: model.var_names_g
          convert_fn: null
```
we can have (less error-prone) configs that look like this
```yaml
    - class_path: cellarium.ml.transforms.ZScore
      init_args: 
        !FileMultiLoader
        file_path: /tmp/test_examples/onepass/onepass.csv
        loader_fn: pandas.read_csv
        fields:
          mean_g:
            attr: mean_g.values
            convert_fn: torch.FloatTensor
          std_g:
            attr: std_g.values
            convert_fn: torch.FloatTensor
          var_names_g:
            attr: var_names_g
            convert_fn: pandas.Series.to_numpy
```

It's still not very concise, but at least we're not having to type the same filename 3 times.

### DIFF
--- a/cellarium/ml/cli.py
+++ b/cellarium/ml/cli.py
@@ -35,7 +35,7 @@ def _resolve_loader(loader_fn: Callable[[str], Any] | str) -> Callable[[str], An
     if isinstance(loader_fn, str):
         loader_fn = import_object(loader_fn)
     if loader_fn not in cached_loaders:
-        cached_loaders[loader_fn] = cache(loader_fn)
+        cached_loaders[loader_fn] = cache(loader_fn)  # type: ignore[arg-type]
     return cached_loaders[loader_fn]
 
 
@@ -52,7 +52,7 @@ def _resolve_value(
     if isinstance(convert_fn, str):
         convert_fn = import_object(convert_fn)
     if convert_fn is not None:
-        obj = convert_fn(obj)
+        obj = convert_fn(obj)  # type: ignore[operator]
     return obj
 
 
--- a/cellarium/ml/cli.py
+++ b/cellarium/ml/cli.py
@@ -105,16 +105,17 @@ class FileMultiLoader:
         model:
           transforms:
             - class_path: cellarium.ml.transforms.ZScore
-              init_args: !FileMultiLoader
-                file_path: gs://dsp-cellarium-cas-public/test-data/onepass.csv
+              init_args:
+                !FileMultiLoader
+                file_path: /tmp/test_examples/onepass/onepass.csv
                 loader_fn: pandas.read_csv
                 fields:
                   mean_g:
-                    attr: mean_g
-                    convert_fn: torch.tensor
+                    attr: mean_g.values
+                    convert_fn: torch.FloatTensor
                   std_g:
-                    attr: std_g
-                    convert_fn: torch.tensor
+                    attr: std_g.values
+                    convert_fn: torch.FloatTensor
                   var_names_g:
                     attr: var_names_g
                     convert_fn: pandas.Series.to_numpy
--- a/cellarium/ml/cli.py
+++ b/cellarium/ml/cli.py
@@ -90,6 +90,81 @@ class FileLoader:
 
 
 @dataclass
+class FileMultiLoader:
+    """
+    A YAML constructor for loading a file once and extracting multiple fields from it.
+
+    Applied to a ``init_args`` mapping, it returns a dict that is unpacked as keyword
+    arguments to the constructor, allowing all fields to be specified in a single block
+    instead of repeating the file path and loader for each argument.
+
+    Example:
+
+    .. code-block:: yaml
+
+        model:
+          transforms:
+            - class_path: cellarium.ml.transforms.ZScore
+              init_args: !FileMultiLoader
+                file_path: gs://dsp-cellarium-cas-public/test-data/onepass.csv
+                loader_fn: pandas.read_csv
+                fields:
+                  mean_g:
+                    attr: mean_g
+                    convert_fn: torch.tensor
+                  std_g:
+                    attr: std_g
+                    convert_fn: torch.tensor
+                  var_names_g:
+                    attr: var_names_g
+                    convert_fn: pandas.Series.to_numpy
+
+    Args:
+        file_path:
+            The file path to load the object from.
+        loader_fn:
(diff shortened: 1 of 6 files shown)


cellarium/ml/cli.py
```diff
@@ -89,6 +89,82 @@ def __new__(cls, file_path, loader_fn, attr=None, key=None, convert_fn=None):
...
+                value = value[key]
+
+            if isinstance(convert_fn, str):
+                convert_fn = import_object(convert_fn)
+            if convert_fn is not None:
+                value = convert_fn(value)
+
+            result[name] = value
+
+        return result
```
fedorgrab: We can factor out some logic, and reuse it in both FileLoader and FileMultiLoader:

```python3
def _resolve_loader(loader_fn: Callable[[str], Any] | str) -> Callable[[str], Any]:
    if isinstance(loader_fn, str):
        loader_fn = import_object(loader_fn)
    if loader_fn not in cached_loaders:
        cached_loaders[loader_fn] = cache(loader_fn)
    return cached_loaders[loader_fn]

def _resolve_value(
    obj: Any,
    attr: str | None = None,
    key: Any = None,
    convert_fn: Callable[[Any], Any] | str | None = None,
) -> Any:
    if attr is not None:
        obj = attrgetter(attr)(obj)
    if key is not None:
        obj = obj[key]
    if isinstance(convert_fn, str):
        convert_fn = import_object(convert_fn)
    if convert_fn is not None:
        obj = convert_fn(obj)
    return obj
```

FileMultiLoader's `__new__` would be like:
```python3
def __new__(cls, file_path, loader_fn, fields) -> dict:  # type: ignore[misc]
    obj = _resolve_loader(loader_fn)(file_path)
    return {name: _resolve_value(obj, **spec) for name, spec in fields.items()}
```

while FileLoader's `__new__` would be like:
```python3
def __new__(cls, file_path, loader_fn, attr=None, key=None, convert_fn=None):
    obj = _resolve_loader(loader_fn)(file_path)
    return _resolve_value(obj, attr=attr, key=key, convert_fn=convert_fn)
```

cellarium/ml/cli.py
```diff
@@ -89,6 +89,82 @@ def __new__(cls, file_path, loader_fn, attr=None, key=None, convert_fn=None):
...
+            A mapping from output key names to field specs. Each spec may contain:
+            ``attr`` (dotted attribute path), ``key`` (item key), and ``convert_fn``
+            (importable callable string or ``None``).
+    """
+
+    file_path: str
+    loader_fn: Callable[[str], Any] | str
+    fields: dict[str, dict]
+
+    def __new__(cls, file_path, loader_fn, fields) -> dict:  # type: ignore[misc]
```
fedorgrab, line 147: type annotations are missing, is this expected?

cellarium/ml/cli.py
```diff
@@ -89,6 +89,82 @@ def __new__(cls, file_path, loader_fn, attr=None, key=None, convert_fn=None):
         return obj
```
fedorgrab: I left a comment about FileMultiLoader which is related to this class too. 

Also, this __new__ method doesn't have type annotations, is this on purpose?

cellarium/ml/cli.py
fedorgrab: Also, maybe worth covering `_resolve_loader` and `_resolve_value` with simple unit tests, if you like this kinda refactor idea.

examples/cli_workflow/ipca_train_config.yaml
```diff
@@ -63,22 +63,20 @@ model:
...
+        fields:
+          mean_g:
+            attr: mean_g.values
+            convert_fn: torch.FloatTensor
+          std_g:
+            attr: std_g.values
+            convert_fn: torch.FloatTensor
+          var_names_g:
+            attr: var_names_g
+            convert_fn: pandas.Series.to_numpy
```
fedorgrab, line 79: That is so much cleaner, I love that!

fedorgrab: How does it know where t get mean_g.values, or std_g.values? shouldn't it be smth like obj.meag_g.values? Or does it apply mean_g to the pandas instance?

fedorgrab (approved):
Looks good – I left a comment with an idea for reducing boilerplate.

examples/cli_workflow/ipca_train_config.yaml
sjfleming: So the `mean_g` in `attr: mean_g.values` refers to the column of the pandas dataframe.  Basically like the obj returned by pandas.read_csv then gets accessed by mean_g, so if it's a column name then it works, and then values further extracts the data as a numpy array.

cellarium/ml/cli.py
sjfleming: Yeah I like this idea

sjfleming: Wow that makes it look so much better!  Nice!

cellarium/ml/cli.py
sjfleming: Yeah the thing is for __new__ it needs to return an instance of itself, but this is kinda being misused in the yaml config context... claude told me to ignore the type annotation stuff around FileLoader and FileMultiLoader

PR #383, opened 2026-05-21
optimize propagation for memory and compilation

sjfleming:
Attempts to optimize the logit probability propagation function to reduce memory use and make things compile nicely with torch.compile

Needs some eyes on it, as it's a lot of AI help


sjfleming:
This was terrible, closing it

sjfleming:
3x slowdown from original, and it was more memory use

PR #384, opened 2026-05-21
remove pyro; subset to active cell types from start not on fly

sjfleming:
Testing out two speed improvements:
- subset `W_gc` and `b_c` to only the cell types used for training before creating those tensors
    - not on the fly during training steps
- get rid of pyro entirely.  seems like it might be causing a lot of cpu overhead


sjfleming:
This ended up being very good.  4x speedup approximately.  I don't know if it was just undoing damage I did before by having a large `W_gc` and indexing it on every forward pass... or if pyro was also introducing a lot of slowness.

Either way, we want these changes

sjfleming:
Actually, now the unconstrained categorical distribution is no longer needed

PR #386, opened 2026-05-22
PyArrow data fomat for speed

sjfleming:
Adds two things:

- dataloader that handles special pyarrow data format built for speed
    - benchmarking makes it seem like 71ms per shard compared to full h5ad at 1222ms or filtered h5ad at 882ms
- a dummy kind of model that is used specifically to take one pass through the data and write this special fast pyarrow file format

Still a bit experimental.  Tests are written but it hasn't been run for real in the wild.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 __pycache__/
-data/
 runs/
 .ipynb_checkpoints
 .mypy_cache/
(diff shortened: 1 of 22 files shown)


sjfleming:
I now think this might be a total rabbit hole.  It seems like PCIe traffic and data volume is the real culprit here, not reading files.  Disk I/O is very low.  Using decompressed h5ad files does not speed things up over using gzip compressed h5ad files.

sjfleming:
This is going to be closed in favor of #387 

It's just not necessary it doesn't seem.  It may give a boost after #387 , but without it, this was no faster.

I may test it again now that #387 exists.  But probably let's just abandon this. It's diminishing returns and pretty complicated.

PR #387, opened 2026-05-23
Dataloading: enable keeping data sparse until gpu

sjfleming:
The idea here is to try to solve the GPU starvation issue.

I've been seeing that it's not file I/O, it's not decompression, it's not python object overhead.  It seems to be PCIe data volume being transferred from the workers to the GPU.  The system just cannot handle the throughput, even an n1-standard-16, not at all.  The issue is that we're passing around dense matrices from the workers.  If we can densify on GPU, we can potentially alleviate this bottleneck.  (Still remains to be proven that this actually makes an epoch faster. EDIT: conclusively proven, it's faster.)

Two new data flow paths:

Scenario | convert_fn | cpu_transforms | first transforms
-- | -- | -- | --
With gene filtering | keep_sparse | Filter(filter_list) | Densify
All genes (OnePass, HVG, …) | to_torch_sparse_csr | (none) | Densify

### DIFF
--- a/cellarium/ml/transforms/densify.py
+++ b/cellarium/ml/transforms/densify.py
@@ -33,7 +33,7 @@ class Densify(nn.Module):
         try:
             if scipy.sparse.issparse(x_ng):
                 raise TypeError(
-                    "Densify received a scipy sparse matrix, which should have been converted to " \
+                    "Densify received a scipy sparse matrix, which should have been converted to "
                     "torch.sparse_csr_tensor earlier in the pipeline. \n\n"
                     "The convert_fn used for x_ng (sparse mode, where Densify is used) must be either:\n"
                     "    - keep_sparse: ensure Filter is in model.cpu_transforms to convert scipy sparse to "
--- a/cellarium/ml/transforms/densify.py
+++ b/cellarium/ml/transforms/densify.py
@@ -1,6 +1,7 @@
 # Copyright Contributors to the Cellarium project.
 # SPDX-License-Identifier: BSD-3-Clause
 
+import scipy.sparse
 import torch
 from torch import nn
 
@@ -24,7 +25,24 @@ class Densify(nn.Module):
 
         Returns:
             A dictionary with the key ``x_ng`` containing a dense :class:`torch.Tensor`.
+
+        Raises:
+            TypeError: If ``x_ng`` is a scipy sparse matrix (which should have been converted earlier).
         """
+        # check for scipy sparse matrices that should have been converted earlier
+        try:
+            if scipy.sparse.issparse(x_ng):
+                raise TypeError(
+                    "Densify received a scipy sparse matrix, which should have been converted to " \
+                    "torch.sparse_csr_tensor earlier in the pipeline. \n\n"
+                    "The convert_fn used for x_ng (sparse mode, where Densify is used) must be either:\n"
+                    "    - keep_sparse: ensure Filter is in model.cpu_transforms to convert scipy sparse to "
+                    "torch.sparse_csr_tensor before GPU transfer.\n"
+                    "    - to_torch_sparse_csr: can be used with no Filter, since CSR is torch.\n"
+                )
+        except TypeError:
+            raise
+
         if x_ng.is_sparse_csr:
             x_ng = x_ng.to_dense()
         return {"x_ng": x_ng}
--- a/cellarium/ml/transforms/densify.py
+++ b/cellarium/ml/transforms/densify.py
@@ -30,18 +30,17 @@ class Densify(nn.Module):
             TypeError: If ``x_ng`` is a scipy sparse matrix (which should have been converted earlier).
         """
         # check for scipy sparse matrices that should have been converted earlier
-        try:
-            if scipy.sparse.issparse(x_ng):
-                raise TypeError(
-                    "Densify received a scipy sparse matrix, which should have been converted to "
-                    "torch.sparse_csr_tensor earlier in the pipeline. \n\n"
-                    "The convert_fn used for x_ng (sparse mode, where Densify is used) must be either:\n"
-                    "    - keep_sparse: ensure Filter is in model.cpu_transforms to convert scipy sparse to "
-                    "torch.sparse_csr_tensor before GPU transfer.\n"
-                    "    - to_torch_sparse_csr: can be used with no Filter, since CSR is torch.\n"
-                )
-        except TypeError:
-            raise
+        if scipy.sparse.issparse(x_ng):
+            raise TypeError(
+                "Densify received a scipy sparse matrix, but sparse inputs should already "
+                "have been converted to torch.sparse_csr_tensor earlier in the pipeline.\n\n"
+                f"Received type: {type(x_ng).__name__}\n\n"
+                "For sparse x_ng with Densify, use one of the following convert_fn options:\n"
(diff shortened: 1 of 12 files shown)


sjfleming:
Experiments on an n1-standard-16 with a single T4 gpu.  Experiments are carried out with 8 dataloader workers and a prefetch_factor of 2 and a batch_size of 4096. (2 min burn-in for tqdm to give a good estimate.)

**For one epoch of the onepass model on 1000 h5ad files with 60k genes (no `Filter`)**

- main branch takes 30 min (gpu util is intermittent)
- this branch takes 12 min (gpu util is high)

**For one epoch of the onepass model on 1000 h5ad files with a `Filter` to 7919 genes**

- main branch with gpu transform takes 26 min (gpu util is intermittent)
- main branch with cpu dataloader transform takes 19 min (gpu util is very intermittent and choppy)
- this branch takes 6 min (gpu util is intermittent and choppy)

and if I take this branch up to 12 dataloader workers (which is no better for the main branch, actually worse):
- this branch takes 5 min (gpu util is intermittent and choppy)
(increasing prefetch_factor to 4 is still 5 min)

and if I take this branch up to 14 dataloader workers with prefetch_factor 4
- this branch takes 4.5 min

and CPU utilization is pretty much maxed out.  GPU util is still choppy, so we're not quite able to feed it fast enough.  But still, big improvement.

sjfleming:
Here's the config file for that last test that did onepass on 1001 h5ad files using 7919 genes in 4 mins 27 sec (full run).

```yaml
seed_everything: true
trainer:
  accelerator: gpu
  strategy: auto
  devices: 1
  num_nodes: 1
  precision: null
  logger: null
  callbacks: null
  fast_dev_run: false
  max_epochs: 1
  min_epochs: null
  max_steps: -1
  min_steps: null
  max_time: null
  limit_train_batches: null
  limit_val_batches: null
  limit_test_batches: null
  limit_predict_batches: null
  overfit_batches: 0.0
  val_check_interval: null
  check_val_every_n_epoch: 1
  num_sanity_val_steps: null
  log_every_n_steps: null
  enable_checkpointing: null
  enable_progress_bar: null
  enable_model_summary: null
  accumulate_grad_batches: 1
  gradient_clip_val: null
  gradient_clip_algorithm: null
  deterministic: null
  benchmark: null
  inference_mode: true
  use_distributed_sampler: true
  profiler: null
  detect_anomaly: false
  barebones: false
  plugins: null
  sync_batchnorm: false
  reload_dataloaders_every_n_epochs: 0
  default_root_dir: null
model:
  cpu_transforms:
    - class_path: cellarium.ml.transforms.Filter
      init_args:
        filter_list:
          !FileLoader
          file_path: dummy_gene_list.csv
          loader_fn: pandas.read_csv
          attr: ensembl_id
          convert_fn: pandas.Series.to_list
  transforms:
    - cellarium.ml.transforms.Densify
    - class_path: cellarium.ml.transforms.NormalizeTotal
      init_args:
        target_count: 10_000
    - cellarium.ml.transforms.Log1p
  model:
    class_path: cellarium.ml.models.OnePassMeanVarStd
    init_args:
      algorithm: shifted_data
      output_path: ./onepass.csv
  optim_fn: null
  optim_kwargs: null
  scheduler_fn: null
  scheduler_kwargs: null
  is_initialized: false
data:
  dadc:
    class_path: cellarium.ml.data.DistributedAnnDataCollection
    init_args:
      filenames: /home/sfleming/data/extract_{000000..001000}.h5ad
      limits: null
      shard_size: 10000
      last_shard_size: null
      max_cache_size: 3
      cache_size_strictly_enforced: true
      label: null
      keys: null
      index_unique: null
      convert: null
      indices_strict: true
      obs_columns_to_validate:
        - raw_sum
  batch_keys:
    x_ng:
      attr: X
      convert_fn: cellarium.ml.utilities.data.keep_sparse
    var_names_g:
      attr: var
      key: feature_id
    total_mrna_umis_n:
      attr: obs
      key: raw_sum
  batch_size: 4096
  iteration_strategy: cache_efficient
  shuffle: true
  shuffle_seed: 0
  drop_last_indices: false
  drop_incomplete_batch: false
  worker_seed: null
  test_mode: false
  num_workers: 14
  prefetch_factor: 4
  persistent_workers: false
ckpt_path: null
```

sjfleming:
Also looked at disk io using
```
iostat -xy 1
```
(this is a networked persistent SSD on a google VM, not actual local SSD)

The usage is a bit high, like 50-80% at times.

I tried
```
sudo blockdev --setra 8192 /dev/root
```

but I'm still seeing 4.5 mins.  I think the `Filter` on scipy sparse on the dataloader workers must be the bottleneck... or just decompressing the files (these are h5ads saved with gzipped count matrix format).

If I try using decompressed h5ads saved with `compression=None`, then I actually see a slowdownb (7min)!  Paradoxical to me.  I guess io is slower actually.  Yes io %util becomes 90+% consistently, that's it.  Disk becomes the bottleneck.

So currently I think the bottleneck is scipy gene filtering.  I wonder how fast this is if we pre-filter the files.

sjfleming:
It's getting a bit hard to mix and match everything in the config in a way that works, and there are a lot of ways to do things wrong.

I'm going to

- [x] add validation and actionable error messages for these patterns
- [x] change the default config files to use this sparse dataloader pattern, since it is so much faster and should be the default mode of operation

sjfleming:
Prefiltering the h5ad files to a few thousand genes speeds things up considerably with this sparse path.  It's like < 2 mins for 1000 h5ad files for onepass.

cellarium/ml/transforms/densify.py
```diff
@@ -0,0 +1,51 @@
...
+                raise TypeError(
+                    "Densify received a scipy sparse matrix, which should have been converted to "
+                    "torch.sparse_csr_tensor earlier in the pipeline. \n\n"
+                    "The convert_fn used for x_ng (sparse mode, where Densify is used) must be either:\n"
+                    "    - keep_sparse: ensure Filter is in model.cpu_transforms to convert scipy sparse to "
+                    "torch.sparse_csr_tensor before GPU transfer.\n"
+                    "    - to_torch_sparse_csr: can be used with no Filter, since CSR is torch.\n"
+                )
+        except TypeError:
+            raise
```
fedorgrab:
```suggestion
        if scipy.sparse.issparse(x_ng):
            raise TypeError(
                "Densify received a scipy sparse matrix, but sparse inputs should already "
                "have been converted to torch.sparse_csr_tensor earlier in the pipeline.\n\n"
                f"Received type: {type(x_ng).__name__}\n\n"
                "For sparse x_ng with Densify, use one of the following convert_fn options:\n"
                "  - keep_sparse: use when Filter is in model.cpu_transforms, so Filter converts "
                "scipy sparse input to torch.sparse_csr_tensor before GPU transfer.\n"
                "  - to_torch_sparse_csr: use when there is no Filter, since the output is already "
                "a torch sparse CSR tensor."
            )
```

fedorgrab (approved):
Left one nit comment, not sure if the nested try catch was intentional.

cellarium/ml/transforms/densify.py
sjfleming: You're right I think I can get rid of that!  I'm also adding a small test for Densify to tests/test_transforms.py

PR #389, opened 2026-06-10
add the kotliar hvg function

sjfleming:
Renames the HVG helper function "get_highly_variable_genes" to "seurat_compute_highly_variable_genes".  Also slightly tweaks the input arg names for standardization.

Adds Kotliar HVG selection based on Fano factors as the function `kotliar_compute_highly_variable_genes`.  (Used in NMF and maybe be useful more broadly.). Adds tests for it.

### DIFF
--- a/cellarium/ml/preprocessing/highly_variable_genes.py
+++ b/cellarium/ml/preprocessing/highly_variable_genes.py
@@ -56,10 +56,10 @@ def _hvg_seurat_single_batch(
     return df
 
 
-def get_highly_variable_genes(
-    gene_names: list,
-    mean: torch.Tensor,
-    var: torch.Tensor,
+def seurat_compute_highly_variable_genes(
+    var_names_g: list,
+    mean_g: torch.Tensor,
+    var_g: torch.Tensor,
     n_top_genes: int | None = None,
     min_disp: float | None = 0.5,
     max_disp: float | None = np.inf,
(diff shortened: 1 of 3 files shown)


cellarium/ml/preprocessing/highly_variable_genes.py
```diff
@@ -56,10 +56,10 @@ def _hvg_seurat_single_batch(
     return df
 
 
-def get_highly_variable_genes(
-    gene_names: list,
-    mean: torch.Tensor,
-    var: torch.Tensor,
+def seurat_compute_highly_variable_genes(
```
fedorgrab, line 59: Thanks for updating this. This naming is consistent with the new code!

cellarium/ml/preprocessing/highly_variable_genes.py
```diff
@@ -261,3 +261,139 @@ def _get_highly_variable_genes_batched(
         )
 
     return df_out
+
+
+def kotliar_compute_highly_variable_genes(
```
fedorgrab, line 266: nit: totally optional, but maybe reordering the input parameters to be consistent with seurat hgvs?

fedorgrab (approved):
I just noticed, seems like we have no words about hvgs in docs? Should we add it? Anyways, that's an open discussion, maybe should be a separate issue!

cellarium/ml/preprocessing/highly_variable_genes.py
sjfleming: thanks yes, good point.  also renamed `num_genes` to `n_top_genes` to match

PR #391, opened 2026-06-16
add CenterPerCell transform and the PFlogPF composition

sjfleming:
Based on the results here

https://www.biorxiv.org/content/10.1101/2022.05.06.490859v3

This transform simply applies a per-cell centering to log1p(fractional_counts) data.

### DIFF
--- a/cellarium/ml/transforms/center_per_cell.py
+++ b/cellarium/ml/transforms/center_per_cell.py
@@ -1,0 +1,22 @@
+# Copyright Contributors to the Cellarium project.
+# SPDX-License-Identifier: BSD-3-Clause
+
+import torch
+
+
+class CenterPerCell(torch.nn.Module):
+    """
+    Center each cell by subtracting its mean across genes.
+
+    NOTE: PFlog1pPF / shifted-CLR-style normalization [1] is:
+        NormalizeTotal + Log1p + CenterPerCell
+
+    References:
+    [1] Booeshaghi, Hallgrimsdottir, Galvez-Merchan, Pachter. Depth normalization for single-cell genomics count data.
+        bioRxiv (2026). https://www.biorxiv.org/content/10.1101/2022.05.06.490859v3
+    """
+
+    def forward(self, x_ng: torch.Tensor) -> dict[str, torch.Tensor]:
+        cell_mean_n1 = x_ng.mean(dim=-1, keepdim=True)
+        x_ng = x_ng - cell_mean_n1
+        return {"x_ng": x_ng}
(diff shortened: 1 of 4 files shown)


cellarium/ml/transforms/center_per_cell.py
```diff
@@ -0,0 +1,22 @@
...
+class CenterPerCell(torch.nn.Module):
+    """
+    Center each cell by subtracting its mean across genes.
+
+    NOTE: PFlog1pPF / shifted-CLR-style normalization [1] is:
+        NormalizeTotal + Log1p + CenterPerCell
+
+    References:
+    [1] Booeshaghi, Hallgrimsdottir, Galvez-Merchan, Pachter. Depth normalization for single-cell genomics count data.
+        bioRxiv (2026). https://www.biorxiv.org/content/10.1101/2022.05.06.490859v3
```
fedorgrab: Do you think we need to include this ref for CenterPerCell? I was thinking centering is a pretty standard operation, not really tight to this paper.

cellarium/ml/transforms/pflogpf.py
```diff
@@ -0,0 +1,32 @@
...
+
+    This is a convenience wrapper for that sequence of transforms, but it provides
+    no additional functionality.
+
+    References:
+    [1] Booeshaghi, Hallgrimsdottir, Galvez-Merchan, Pachter. Depth normalization for single-cell genomics count data.
+        bioRxiv (2026). https://www.biorxiv.org/content/10.1101/2022.05.06.490859v3
+    """
+
+    def __init__(self, target_count: int = 1):
```
fedorgrab: I would make `target_count`'s default to be `10_000` to be consistent with the default value of NormalizeTotal, or is this this for a reason set to 1?

fedorgrab: Maybe we also need to expose eps for NormalizeTotal, even tho it's unlikely anyone would need to set a different value, but to have a better modular API, I would expose eps and put the same default value to it as in normalize total.

fedorgrab (approved):
I added couple of comments, I would consider them "nit"s. Looks very good, excited to see it in action!

cellarium/ml/transforms/center_per_cell.py
sjfleming: you might be right, yeah it's not something they invented... let's leave the reference in PFlogPF only

cellarium/ml/transforms/pflogpf.py
sjfleming: Actually I had `target_count` as 10k at first, but I realized their formula in PFlogPF uses `target_count` 1.  It's just straight up normalization, not scaling to a new count value.

Okay yeah I'll expose `eps`

fedorgrab: Oh got it, okay then! I just thought was unintentional, if that's how they do it in paper, then makes sense!

PR #395, opened 2026-06-24
patch EarlyStopping and scvi: fix validation metric logging

sjfleming:
EarlyStopping cannot see metrics logged using `trainer.logger.log_metrics`

The right way seems to be `lightning_module.log` so the metrics become visible.

This PR changes the call signature for `on_validation_epoch_end` in models.  Currently scvi is the only model using this hook.

Also important:

`EarlyStopping` is not really compatible with cellarium-ml directly.  cellarium-ml initializes everything on meta device, and when this is happening, EarlyStopping gets instantiated.  It pulls device context from the init context.  So it initializes all its internal tracking stuff on the meta device.  This then errors during training.

This patch creates `MetaSafeEarlyStopping` which just forces the internal trackers onto cpu (which can then get cast correctly later).

### DIFF
--- a/cellarium/ml/callbacks/__init__.py
+++ b/cellarium/ml/callbacks/__init__.py
@@ -2,9 +2,17 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 from cellarium.ml.callbacks.compute_norm import ComputeNorm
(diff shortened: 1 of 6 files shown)


fedorgrab (approved):
Looks very clean!

sjfleming:
Note to self: maybe move scvi to single device testing only, it’s timing out for some reason

PR #396, opened 2026-06-30
Workaround argument linking in CLI for scvi

sjfleming:
With a trained scvi model running `predict`, argument linking in the lightning CLI enforces that things like `n_batch` and `n_cats_per_cov` come from the data.  This is very convenient during training, since the user doesn't have to manually specify everything (error prone).  But during prediction, it's possible `n_batch` and `n_cats_per_cov` are not used (in the encoder).  If the model is approached with new data (which is a common use case), then the lightning CLI will fail on instantiation.  Series of events:

- lightning instantiates model
- argument linking sets `n_batch` based on dataset
- lightning tries to load state dict from the saved checkpoint
- state dict loading fails due to tensor shapes mismatch (different `n_batch` during training made tensor shapes different)
- cannot proceed

This patch adds a specific environment variable to omit that automatic argument linking.

### DIFF
--- a/cellarium/ml/cli.py
+++ b/cellarium/ml/cli.py
@@ -6,6 +6,7 @@ Command line interface for Cellarium ML.
 """
 
 import copy
+import os
 import sys
 import warnings
 from collections.abc import Callable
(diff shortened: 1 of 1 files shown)


sjfleming:
So this is purely a work-around for the specific case of training an scVI model and then wanting to run `predict` on the trained model... AND your model does not use batch covariates in the encoder (thus `predict` does not use them).  You can run

```bash
SCVI_PREDICT_SKIP_ARG_LINKING=1 cellarium-ml scvi predict -c predict_config.yaml
```

fedorgrab (approved):
I tried training a model with this, it works!

sjfleming:
I'm gonna merge this and leave the muP test issue for a later issue

PR #406, opened 2026-07-09
socam: absolute loss scaling

### LINKED ISSUE #405: socam: absolute value of loss is huge
This is because when we went from pyro to pytorch, we scaled up the cross entropy loss rather than scaling down the prior loss.  What's important for correctness is the ratio between the two.

The absolute loss value doesn't matter for Adam, but it matters for SGD and AdamW.

The thing would be to change this
https://github.com/cellarium-ai/cellarium-ml/blob/9f185c7daed8176023351c10bae1d16b930ca1d5/cellarium/ml/models/socam.py#L310-L313

to this
```python
 ce_loss = torch.nn.functional.cross_entropy(logits_nc, y_n, reduction="mean")
 laplace_loss = (self.W_gc.abs().sum() / self.W_prior_scale) / self.n_obs
 loss = ce_loss + laplace_loss 
```

sjfleming:
Closes #405


fedorgrab (approved):
lgtm!

PR #414, opened 2026-07-17
scvi: implement SwiGLU and ResNet as options

sjfleming:
When training scVI at "foundation model" scale (i.e. all of cellxgene), it might be beneficial to have the option to beef up the capacity of the encoder and decoder networks.  They do have to learn 10k+ batch effects after all.  Typically vanilla MLPs are fine for several-dataset level integration that people typically do with scvi-tools.  But the kind of thing we're doing may actually benefit from more capacity.  Remains to be seen.

Anyway, this implements `SwiGLUActivation` (which is a real cool, powerful activation function), and the ResNet block idea, where `DressedLayer` dressing is allowed to add the input back to the output of a same-size-input-and-output layer.  This might allow for constant-width layers 5 or 6 blocks deep.  Again, remains to be seen if scVI likes it.

The SwiGLU activation function is weird because it halves the feature size when it acts.  So this has to be taken into account at layer shape construction size, and actually the thing right before the SwiGLU gets automatically doubled in out_features size, so that when SwiGLU halves it, we get what we wanted.

### DIFF
--- a/cellarium/ml/layers/__init__.py
+++ b/cellarium/ml/layers/__init__.py
@@ -8,11 +8,12 @@ from cellarium.ml.layers.head import MultiHeadReadout
 from cellarium.ml.layers.mu_linear import MuLinear
 from cellarium.ml.layers.normadd import NormAdd
 from cellarium.ml.layers.transformer import Transformer, TransformerBlock
-from cellarium.ml.layers.vae import DressedLayer, FullyConnectedLinear
+from cellarium.ml.layers.vae import DressedLayer, FullyConnectedLinear, SwiGLUActivation
 
 __all__ = [
     "DressedLayer",
     "FullyConnectedLinear",
+    "SwiGLUActivation",
     "TokenEmbedding",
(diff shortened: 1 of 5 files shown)


sjfleming:
Note about SwiGLU as implemented here: sometimes it's implemented as a Module itself, with learnable parameters for two separate linear transformations.  Here, we assume the linear layer has been specified before the SwiGLU (and its feature size doubled), and here the SwiGLU is treated solely as a parameter-free activation function that halves the feature dimension.  That way SwiGLU can become part of a `DressedLayer` as the dressing, i.e. a pure activation function (though it does halve the feature dimension of whatever comes right before it, so this is taken into account when sizing the linear layers).  This integrates it into the current code architecture with minimal changes, and importantly, allows batch information to be injected into both parts of the SwiGLU (gate and values), which allows the network to learn in a batch-dependent manner which features to pass and which to suppress.

fedorgrab (approved):

PR #415, opened 2026-07-21
scANVI

sjfleming:
Implement the scANVI model
https://docs.scvi-tools.org/en/stable/user_guide/models/scanvi.html

scANVI is a super cool model for cell type annotation that was published by Xu et al in 2021, and is based on the M1+M2 model from Kingma's 2014 "semi-supervised deep learning with deep generative models".

This implementation re-uses as much of the scVI implementation as possible.

In the future it may be possible to extend this in two very interesting ways for CAS:
1. calibrate the uncertainties in cell type label prediction to be accurate
2. compute an estimate for out-of-distribution probability, i.e. the probability that an observation is "novel"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ output/
 *.tsv.gz
 *.ckpt
 CLAUDE.md
+benchmarking/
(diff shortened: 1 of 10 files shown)


sjfleming:
This model has several really cool advances beyond what scANVI does.  Firstly, it uses the cell ontology hierarchy itself as a map for label uncertainty marginalization (i.e. coarse author labels). This is such an interesting idea. Hopefully this will enable us to find a totally self-consistent representation of the cells regardless of a given author's chosen label granularity level. Secondly, it uses SOCAM as its classifier, with all of SOCAM's benefits like taking the cell ontology into account in scoring label accuracy (can also do class balancing).

PR #416, opened 2026-07-22
socam: reduce memory and improve numerical stability

sjfleming:
Revamp the computation for probability propagation.

This should improve numerical stability and reduce memory usage while giving the same output.  Works better compiled.


fedorgrab (approved):
lgtm

PR #421, opened 2026-09-01
NMF amortized: BilinearLoadingsEncoder rather than FiLM MLP

sjfleming:
Attempt to improve the hot-start encoder.  The previous FiLM MLP had major deficiencies and mostly acted as a parametrizable, fixed-memory lookup table.  This new `BilinearLoadingsEncoder` does something more intelligent to try to use the data and the current factors to approximate the loadings.  It also keeps replicates more independent than FiLM did. 

As a corollary, I think we have a better stopping criterion as well: the behavior of the encoder's loss.  When the factors are a moving target, the encoder is unable to catch up.  It's only when the factors settle that this loss really settles.  This is a pretty smooth stopping criterion and seems to work well.

Also fixes an issue in the core: `CellariumModule` was not correctly treating buffers attached to `Transforms` that were specified as CPU transforms.  For instance, scale from `DivideByScale` as a cpu_transform would end up with its parameters on GPU if the module was moved to GPU.  This PR fixes that.


sjfleming:
image

Automatic stopping using the new encoder loss.  Each data fold cNMF run takes approx 3 mins on an M2 mac using MPS with compiled FISTA.

sjfleming:
Added a two-part stopping condition.  The encoder loss stopping was not sufficient for another dataset.  So I think a combination of demanding both "encoder loss is not improving" and "reconstruction error is not improving" should work in general.  This is the result on the 4 folds data benchmark using the new stopping criteria:

image

Updated the Haley notebook to use the amortized model.  Trains 180k cell dataset on k in [3, 30] in 196 mins on an M2 Mac laptop, using MPS during training.  The consensus operation and all related plotting take < 10 mins. Here is an example of the Kotliar stability plot:

image

PR #423, opened 2026-09-21
scANVI: return un-propagated raw logits

sjfleming:
Have `predict()` return the raw logits as `"cell_logits_nc"` and changed SOCAM to use the same name.  This is for compatibility with CAS, so CAS can know what to expect.


sjfleming:
I think I'm just gonna close this in favor of #424 , since that includes these changes

PR #424, opened 2026-09-21
Refactor typing for predictions

sjfleming:
Replaces #423 

In the past, a `CellariumModel` was denoted as being able to make predictions by use of inheritance of `PredictMixin`.

This works fine, so authors of models can state explicitly whether the model needs to implement the predict method.

But typing is a bit tricky and has been a free-for-all.  Models can predict whatever they want.

Now, models can still predict whatever they want, but new type annotations are included to allow some `PredictMixin` models to have a more explicit contract about what they return.  For example, a `TransformPrediction` includes the `"x_ng"` key (and `"var_names_g"`) so that a model checkpoint can be used as a `Transform` in cellarium-ml.  And a `ClassifierPrediction` has the keys `"y_logits_nc"` and `"y_probs_nc"` and `"category_labels_c"`. A `TransformClassifierPrediction` is a combo of the other two.

Models have been updated to adhere to this, and in some cases additional outputs for `predict()` have been renamed or added where before they were missing.

I think this helps make our intent about `predict()` methods more explicit.


cellarium/ml/models/model.py
```diff
@@ -74,13 +74,39 @@ def __delattr__(self, name: str) -> None:
...
+    cell_type_logits_nc: torch.Tensor
+    cell_type_probs_nc: torch.Tensor
+
+
+class CASPrediction(TransformPrediction):
+    """Prediction for CAS models, extending the TransformPrediction with cell type logits
+    and probabilities using the keys CAS expects."""
+
+    cell_type_logits_nc: torch.Tensor
+    cell_type_probs_nc: torch.Tensor
```
fedorgrab: This is super cool! One thing is, I wonder if CAS is a right word here in naming? To me it seems that here it'd be better to be cellairum-ml first. This way CAS would be adopting to the framework. I don't think it matters that much, but tell me if you think this vision makes sense? 

If you like it, this could be: ModelPrediction, ClassificationModelPrediction or ClassifierPrediction (when imported `from cellarium.ml.models.model import ClassifierPrediction`).

fedorgrab: Also, just noticed – CASClassifierPrediction and CASPrediction seem to be the same

fedorgrab (changes requested):
This seems to be a very good change, I like it! I left 2 comments, one nit, and another one was about two classes having the same body, not sure if it was intentional.

cellarium/ml/models/model.py
fedorgrab: One more thing. Should we think label name agnostic? If we decide to train SOCAM to predict disease, we would need to go for cell_type_logits_nc to get the prediction (this will be a breaking change for CAS's implementation of SOCAM serving, but it's not a problem).

fedorgrab: I reread the general description of this PR and now I understand that CASPrediction and CASClassifierPrediction are different!! My bad, I didn't notice this at first! 

How about changing the definition of CASPrediction to be this:
```python3
class CASPrediction(TransformPrediction, CASClassifierPrediction):
    """Prediction for CAS models, extending the TransformPrediction with cell type logits
    and probabilities using the keys CAS expects."""
```

This way it's more intuitive (imho) to read. TypedDict supports multiple inheritance, so `CASPrediction` can simply combine the fields from both parent types.

sjfleming: Okay I think these are all good ideas.
- take "CAS" out of the naming
- go back to "y_" instead of "cell_type_", because as you rightly point out, it could be something other than cell type.  although I'm not sure if disease will actually work.  I'm afraid SOCAM might break with some other ontology, but maybe it'll just work.
- clarify the difference between the combo predictor and the classifier only predictor in the definition

sjfleming:
Okay I have made these changes and also added a required predict output key called "category_label_c" for classifiers.

fedorgrab (approved):
LGTM!

